PR #5, opened 2017-01-27
Initial WIP for procbots.

hedss:
See README.md for details.


hedss:
All,
The code is not yet 'complete' in that, whilst functionally it works correctly on repositories, the ProcBots themselves need logging/alerting adding,  @brownjohnf has very kindly offered to help me with this, so the plan is to hand this over to him. ;-) So whilst this is indeed a review, it's a tentative review because of this and especially because...

This is my first major foray into TypeScript:
@emirotin I think you were working on a coding standard, is this ready yet?
@pimterry You have probably come across this, what's the best format for type declaration for things such as wrapping modules in Promises (specifically Bluebird), without throwing this info away?

For clarity, I've attempted to defined partial interfaces for those components used heavily (such as the `github` NPM module, which doesn't have interfaces in its `@types` definition) and generally for anything created by a class.

I've built a `gulpfile` configuration that seems to work. I've come across a few oddities that need specifically adding to allow mapping to work in VS Code. These are documented in the relevant places.

I've also carried out a possible JSDoc implementation for the Worker class in ProcBots.

I'd appreciate any thoughts on any of this. Cheers! :)

package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "resin-procbots",
+  "version": "0.0.1",
+  "description": "Process robots for automating CI pipelines",
+  "homepage": "https://github.com/resin-io-modules/resin-procbots#readme",
+  "preferGlobal": true,
+  "types": "lib/**/*.ts",
```
pimterry: I think (though I'm not 100% sure) `types` is supposed to point to a compiled set of type definitions, rather than the an entire typed implementation. That's certainly far more common. See [resin-sdk-preconfigured](https://github.com/resin-io-modules/resin-sdk-preconfigured) for example, which outputs `.js`, `.js.map` and `.d.ts` (just set `declaration: true` in tsconfig) into `/build`

tools/initRepo/lib/initRepo.ts
```diff
@@ -0,0 +1,169 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Import most things with types if possible.
+import Opts = require('node-getopt');
+import * as _ from 'lodash';
+import * as Promise from 'bluebird';
+
+const hmac = require('crypto');
```
pimterry: Here too, `crypto` should definitely have good working type definitions.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+import * as _ from 'lodash';
+import * as fs from 'fs';
+import * as jwt from 'jsonwebtoken';
+import { AlertLevel } from './procbot';
+
+// The GithubAPI Promise implies built-in ES6 Promises. Therefore when assigning
+// the Promise to use as Bluebird, VS Code reports an error. Need a good
+// fix for this that doesn't require type declarations?
+const GithubApi = require('github');
+const request: any = Promise.promisifyAll(require('request'));
```
pimterry: https://www.npmjs.com/package/request-promise - Bluebird built in (and other builds available for diff Promises if you like), with type definitions published too.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+    private user: number;
+    private eventTriggers: GithubActionRegister[] = [];
+    protected githubApi: any;
+
+    // Takes a set of webhook types that the bot is interested in.
+    // Registrations can be passed in on bot creation, or registered/deregistered later.
+    // However, 'baked in' registrations are not available for deregistration.
+    constructor(integration: number) {
+        super();
+        this._botname = 'GithubBot';
```
pimterry: The setting of `_botname` feels like a strangely secret interface of ProcBot. Why not make it an optional constructor argument, so you can just pass it to `super()`?

lib/versionbot.ts
```diff
@@ -0,0 +1,479 @@
...
+            }).then((headCommit: GithubBot.Commit) => {
+                // We will go ahead and perform a merge if we see Versionbot has committed
+                // something with 'CHANGELOG.md' in it.
+                const commit = headCommit.commit;
+                const files = headCommit.files;
+
+                if ((commit.committer.name === process.env.VERSIONBOT_NAME) &&
+                    _.find(files, (file: GithubBot.CommitFile) => {
+                        return file.filename === 'CHANGELOG.md';
+                    })) {
```
pimterry: The indentation here plus the double indentation of the block below and the callback inline in the test makes this really really confusing to read, for me at least.

lib/versionbot.ts
```diff
@@ -0,0 +1,479 @@
...
+                        // We go ahead and merge.
+                        const commitVersion = commit.message;
+                        return this.gitCall(githubApi.pullRequests.merge, {
+                            owner,
+                            repo: name,
+                            number: pr.number,
+                            commit_title: `Auto-merge for PR ${pr.number} via Versionbot`
+                        }, 3).then((mergedData:GithubBot. Merge) => {
```
pimterry: This triple nested level of promises in this fairly hefty function is all a bit hard to keep track of. Can we break it up a bit?

You could flatten easily by returning immediately from the `gitCall` for `getCommit` above, and this one here is trickier, you could pull this chunk out nicely into a separate `merge` function:

```
if (committer === versionbot && ChangelogWasCommitted) {
   return merge(pr, owner, name, headCommit);
}
```

pimterry: Bonus nit: this line has a space in the argument's type? I'm surprised that compiles, strange.

lib/versionbot.ts
```diff
@@ -0,0 +1,479 @@
...
+            //            approved = false;
+            //      }
+            //     });
+            //
+            // Because of the CFA going on here: https://github.com/Microsoft/TypeScript/issues/11498
+            // This errors because it's optimistic that [].forEach() execution can't affect the
+            // 'approved' variable, therefore it implicitly believes it can only ever be 'false'
+            // and barfs on the 'if (approved === true)' line claiming 'true' cannot be 'false'.
+            // Not great. Still, easy to use a straight for() loop.
+            approved = false;
```
pimterry: You've got `approved = false` here twice.

Also, I think you can get out of the issues you explain in the comment above if you assert `approved` as boolean in the if condition. Not sure if it's significantly better, but an interesting option:

```typescript
if (approved === true) {
  // Doesn't compile, === can't be applied to false and true
}

if (<boolean> approved === true) {
  // Works fine
}
```

lib/app.ts
```diff
@@ -0,0 +1,102 @@
...
+	if (!verifyWebhookToken(JSON.stringify(payload), req.get('x-hub-signature'))) {
+		res.sendStatus(401);
+		return;
+	}
+
+	// Let the hook get on with it.
+	res.sendStatus(200);
+
+	// Go through all registered bots, and send them any appropriate hook.
+	_.forEach(botRegistry, (bot: GithubBot.GithubBot) => {
```
CameronDiver: +1 for the explicitness IMO, I mean that's why we're using typescript in the first place. The types don't exist at interpreter level which means that they are only there for the programmer and compiler, so whilst it can be more helpful to let it infer, I think in the interest of understanding being explicit is better.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+// These provide the current bare minimum definitions for child Procbots working with them.
+export interface CommitFile {
+    filename: string,
+};
+
+export interface Commit {
+    commit: {
+        committer: {
+            name: string
+        }
```
CameronDiver: Perhaps email as well? The procbot may need to alert the committer in some way that action is required.

lib/app.ts
```diff
@@ -0,0 +1,102 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Import most things with types if possible.
+import Opts = require('node-getopt');
```
hedss: Pure forgetfulness that I'd included them, had meant to change it prior to release.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// The Register interface is passed to the GithubBot.register method to register
+// for callback when the appropriate events and labels are received.
+export interface GithubActionRegister extends GithubAction {
+    workerMethod: GithubActionMethod
+}
+
+// A GithubActionMethod is the method that will be used to process an event.
+export type GithubActionMethod = <T>(action: GithubAction, data: T) => Promise<void>;
```
CameronDiver: Agreed, keep the model and the logic seperate as much as possible.

lib/app.ts
```diff
@@ -0,0 +1,102 @@
...
+		console.log(`Could not import bot type ${bot}`);
+		console.log(err);
+		throw err;
+	}
+}
+
+// Run the hook listener now, and for each bot we've been asked to listen
+// to, wait for events. When we get an event, check against those bots
+// registered, and send filtered messages on to them for action.
+app.post('/webhooks', (req: any, res: any) => {
```
hedss: I've left some of these in for commenting on; I'm more than happy to rely on inferring all types, though I did wonder if it's sometimes clearer to ensure. In the case of `any`, I guess it really doesn't help.

lib/app.ts
pimterry: You can still get the info by mousing over in VS Code, if that's what you're looking for.

Regardless though, sure, if people are getting value from this being more explicit then that's good enough for me :+1:

hedss: I'd left it to ensure that it was realised that it's firing on the inherited class. But actually, now I look at it again, the type of `botRegistry` isn't too far up and is explicit, so I think I'll remove it.

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
```
CameronDiver: We probably want `declarations: true` here so that typescript will generate type information for anyone using this module.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import * as fs from 'fs';
+import * as jwt from 'jsonwebtoken';
+import { AlertLevel } from './procbot';
+
+// The GithubAPI Promise implies built-in ES6 Promises. Therefore when assigning
+// the Promise to use as Bluebird, VS Code reports an error. Need a good
+// fix for this that doesn't require type declarations?
+const GithubApi = require('github');
```
hedss: So to be clearer, there are `@type` declarations for `github`, but importing with:
```
import * as GithubApi from 'github';
```
 fails. This, I think, is due to the constructor itself takes a 'Promise' property (see line 204 of `githubbot.ts`):
```
lib/githubbot.ts(198,40): error TS2345: Argument of type '{ protocol: string; host: string; headers: { 'Accept': string; }; Promise: typeof Bluebird; timeo...' is not assignable to parameter of type 'Options | undefined'.
  Type '{ protocol: string; host: string; headers: { 'Accept': string; }; Promise: typeof Bluebird; timeo...' is not assignable to type 'Options'.
    Types of property 'Promise' are incompatible.
      Type 'typeof Bluebird' is not assignable to type 'PromiseConstructor | undefined'.
        Type 'typeof Bluebird' is not assignable to type 'PromiseConstructor'.
          Types of property 'all' are incompatible.
            Type '{ <T1, T2, T3, T4, T5>(values: [T1 | Thenable<T1>, T2 | Thenable<T2>, T3 | Thenable<T3>, T4 | The...' is not assignable to type '{ <TAll>(values: Iterable<TAll | PromiseLike<TAll>>): Promise<TAll[]>; <T1, T2, T3, T4, T5, T6, T...'.
              Type 'Bluebird<[any, any, any, any, any]>' is not assignable to type 'Promise<any[]>'.
                Types of property 'then' are incompatible.
                  Type '{ <U1, U2>(onFulfill: (value: [any, any, any, any, any]) => U1 | Thenable<U1>, onReject: (error: ...' is not assignable to type '{ (onfulfilled?: ((value: any[]) => any[] | PromiseLike<any[]>) | null | undefined, onrejected?: ...'.
                    Type 'Bluebird<any>' is not assignable to type 'Promise<any[]>'.
                      Types of property 'then' are incompatible.
                        Type '{ <U1, U2>(onFulfill: (value: any) => U1 | Thenable<U1>, onReject: (error: any) => U2 | Thenable<...' is not assignable to type '{ (onfulfilled?: ((value: any[]) => any[] | PromiseLike<any[]>) | null | undefined, onrejected?: ...'.
                          Type 'Bluebird<any>' is not assignable to type 'Promise<any[]>'.
```

lib/githubbot.ts
hedss: Thanks, not sure why I hadn't actually searched for this!

lib/githubbot.ts
hedss: I originally had... and I'm not quite sure why it's changed. I suspect during other changes, and I just didn't revert this.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+            this.log(ProcBot.LogLevel.DEBUG, 'Base curl command:');
+            this.log(ProcBot.LogLevel.DEBUG, `curl -XGET -H "Authorisation: token ${tokenDetails.token}" -H "Accept: application/vnd.github.black-cat-preview+json" https://api.github.com/`)
+        });
+    }
+
+    // Make a 'github' API call. We explicitly wrap this so that any authentication error
+    // can result in re-authentication before moving on.
+    protected gitCall = (method: any, options: any, retries?: number): Promise<any> => {
+        let badCreds = false;
+        let retriesLeft: number = retries || 3;
```
hedss: Yep, as above, going to assume everything is now inferred where possible.

lib/githubbot.ts
CameronDiver: Not sure if this is helpful in this situation or not, but how I used the promise types is like this: https://github.com/resin-io/resin-docker-build/pull/1/files#diff-f41e9d04a45c83f3b6f6e630f10117feR11 and https://github.com/resin-io/resin-docker-build/pull/1/files#diff-f41e9d04a45c83f3b6f6e630f10117feR178 but I'm also v. unfamiliar with ts so could be wrong in my usage.

lib/githubbot.ts
```diff
@@ -0,0 +1,403 @@
...
+
+            // Kick it off.
+            runApi();
+        });
+    }
+}
+
+// Create a new GithubBot. Or rather, don't.
+export function createBot(): GithubBot {
+    throw new Error('GithubBot is not a valid instantiation of a ProcBot');
```
hedss: I was on the fence about this. I've followed a few C++ patterns here, as it's the closest reference I have to a language I'm comfortable with with many of the same concepts. Including a factory that shouldn't be called would make a clear statement that it's essentially a virtual class.. but this doesn't actually make sense in TS. Shall remove.

tools/initRepo/lib/initRepo.ts
hedss: Yep, and in fact in `app.ts` I use it. Purely forgetfulness again.

lib/versionbot.ts
hedss: Shall change.

lib/versionbot.ts
hedss: Yep, in fact it's been my intention to do this, but I really wanted some initial comments on this before the weekend. :)

lib/versionbot.ts
hedss: Ah, explicit casting! Cool, thanks.

lib/versionbot.ts
CameronDiver: It looks like valid syntax to me. Whitespace shouldn't make a difference.

lib/procbot.ts
```diff
@@ -0,0 +1,265 @@
...
+        .then(() => {
+            if (this.queue.length > 0) {
+                process.nextTick(this.runWorker);
+            } else {
+                // Unlink ourselves from our parent list.
+                self.parentMap.delete(self.context);
+            }
+        });
+    }
+}
```
hedss: OK, this is where my C++ism really kicked in. TS doesn't let you (AFAIK, having read around) employ nested classes, which is how I'd have tacked this originally, with parts of the class as protected and all others private.

My issue here is that whilst some of the Worker types are to be exported, some (such as WorkerMap for example), I'd rather keep module private. If I split this up into a separate Worker file, I'd have to expose everything, right? Is there a good paradigm for doing this that isn't class inheritance, for example? (And hopefully that all makes sense!)

package.json
hedss: Agree, and it's what I did for using `resin-discoverable-services`. I'm not sure why I've done this.

lib/githubbot.ts
hedss: See comment below; essentially exposing all interfaces, even when some are only going to be used by a specific class 'seems' wrong.

lib/githubbot.ts
pimterry: If you're looking for purely virtual classes, there's an actual `abstract` keyword for that. `abstract class` will give you a class you can't directly instantiate, but can extend. I think that's what you're looking for?

lib/app.ts
pimterry: Ok, it sounds like for code like this, if you like it, you should leave them in.

For things like `var x: X = new X()`, where the type is clear and very close to the definition, definitely drop it.

In the middle ground, pick whichever option is closer. Here though I think you're right @CameronDiver that it's not instantly obvious what the type is - it's a couple of blocks back and you can't immediately see it at a glance - so this seems reasonably useful to me.

lib/githubbot.ts
pimterry: I did some more investigation. The deep deep problem is that a Bluebird promise instance doesn't have the same type as a full real-world promise instance, because it doesn't have a `[Symbol.toStringTag]` key ([bluebird bug](https://github.com/petkaantonov/bluebird/issues/1277), [corresponding type definitions bug](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/11027#issuecomment-248642717)).

The Github types are determinedly asking for something with exactly the same type as a real implementation (`typeof Promise`), not just something that merely fulfills the spec, so the error's actually arguably right: bluebird isn't sufficient for what the github types are asking for. If, hypothetically, github did expect that symbol key, it'd break. Really, github should probably be asking for a `PromiseConstructorLike`, which is a way looser interface that just says you look like something that probably follows the spec.

It doesn't do that though, and it seems like there's other people looking at proper solutions on the Bluebird side, so for now I'd just ignore it. If you use proper types here, but use `Promise: <any> Promise` in the github call on line 204, I think everything else will keep typing happily, and you can skip merrily past the problem.

tools/initRepo/lib/initRepo.ts
pimterry: There's linting checks that will catch this in ts-lint I think. Might be time to come up with a standard tsconfig and linting setup soon...

lib/procbot.ts
pimterry: You're right, TypeScript doesn't really have a concept of module privacy - everything is file scoped, and you can't have something (afaik) that's shared between multiple files but otherwise private without doing some very hacky things with inheritance, which I'd avoid.

I think the prevailing logic is that this is fine though. If you have two entities that are meaningful standalone, and they share a common model, then that model is reasonably meaningful standalone in your codebase too, and should be public. Since Worker is an exported class and this is the type of its constructor argument, WorkerMap is more or less part of your domain anyway. Alternatively, you're sharing an implementation detail here that other classes shouldn't even know about, and it actually shouldn't be shared at all.

Some interesting solutions:

* Don't tell Worker about the WorkerMap. It currently only uses it to remove itself from it, but arguably that's it having too much knowledge about how it's used. You could just pass an `onDone` callback in when it's created, which calls some `.removeWorker` method to unregister itself. Then all it knows about is the callback it's given, the procbot subclasses just know about the add/get/remove worker methods, and ProcBot itself is the only one who knows there's a map involved. **I think this is actually probably the right option here.**
* Just make more things public and don't worry about it.
* Make things public, but use the structure to make things clearer. You could have lib/procbot/procbot.ts, lib/procbot/worker.ts, lib/procbot/internals.ts, which clearly points people to the right bits of the code, but doesn't require you to bother enforcing it.
* Duplicate type between the Worker and ProcBot modules. Feels nasty, but there's no concrete downside I think, since structural typing means they'll still be considered the same, and type checking will stop you ever going wrong. Also lets them evolve independently - you can have extra information in the map that ProcBot uses, without Workers finding out about it.

pimterry: Oh also you can have nested classes, it's just they're a bit weird: http://stackoverflow.com/a/21920276/68051

lib/githubbot.ts
pimterry: I think for this bit, they should indeed definitely all be public, in a separate file. They're all already exported publically anyway, and you'll need them in other downstream Github bots, so they're a meaningful part of the interface.

lib/procbot.ts
hedss: I agree with the first option, a callback would be the right way to go about this (the joys of prototyping into production).

Declaring a module inside a class. That's not a massive hack to get round JavaScript restrictions, is it? ;-)

lib/app.ts
hedss: Actually, I now remember why. It's because the `node-getopt` declarations declare using the `export = ` format and not a namespace. Therefore attempting to 'import ... from 'node-getopt'` fails, as it doesn't resolve to a module. Keeping as-is.

pimterry: Ah, makes sense. I think maybe I'd parsed this as `var Opts = require('node-getopt')`, which is the case we need to avoid. `import X = ...` I'm fine with, if that's what works.

emirotin:
I didn't do anything for the TS coding standard TBH. If @pimterry can recommend any popular linting preset I'd be fine with it without much bikeshedding.

hedss:
@emirotin No worries, we've had some discussion and I'm going to put something together in the next day as a spec. we can all discuss. :)

tsconfig.json
hedss: Done.

tools/initRepo/lib/initRepo.ts
```diff
@@ -0,0 +1,169 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Import most things with types if possible.
+import Opts = require('node-getopt');
```
hedss: As with comment above, this is required.

lib/versionbot.ts
```diff
@@ -0,0 +1,479 @@
...
+    file: string,
+    encoding: string
+}
+
+interface EncodedFile extends FileMapping {
+    treeEntry: GithubBot.TreeEntry,
+    blobSha: string
+};
+
+// The VersionBot is built ontop of GithubBot, which does all the heavy lifting and scheduling.
```
hedss: Done

tslint.json
```diff
@@ -0,0 +1,25 @@
+{
+    "extends": "tslint:recommended",
+    "rules": {
+        "indent": "tabs",
```
CameronDiver: Didn't we agree on spaces?

tsconfig.json
```diff
@@ -1,15 +1,16 @@
...
+    "compilerOptions": {
+        "module": "commonjs",
+        "declaration": true,
+        "noImplicitAny": true,
+        "removeComments": true,
+        "preserveConstEnums": true,
+        "strictNullChecks": true,
+        "sourceMap": true,
+        "target": "es6",
+        "baseUrl": "./typings",
```
pimterry: I'm surprised by `baseUrl`. I've never seen it before, and [the docs](https://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url) say it exists mainly for AMD projects.

Is this so that `require('temp')` finds `typings/temp.d.ts`? In that case, I think you just need to ensure that the file is already included in the build process - i.e. add `typings/*.d.ts` to `includes` here or your `gulp.files()` argument above. The module shouldn't need to resolved correctly from its name because it's an ambient module declaration - it will define the module name you can load it by explicitly by itself, independent of its location, as long as it's in the build.

lib/versionbot.ts
```diff
@@ -25,14 +25,14 @@ import * as GithubBotApiTypes from './githubapi-types';
...
 // Exec technically has a binding because of it's Node typings, but promisify doesn't include
-// the optional object (we need for CWD).
-const fsReadFile = Promise.promisify(FS.readFile);
+// the optional object (we need for CWD). So we need to special case it.
 const exec: (command: string, options?: any) => Promise<{}> = Promise.promisify(ChildProcess.exec);
-const tempMkdir: (dir: string) => Promise<{}> = Promise.promisify(require('temp').mkdir);
-const tempCleanup = Promise.promisify(require('temp').cleanup);
+const fsReadFile = Promise.promisify(FS.readFile);
+const tempMkdir = Promise.promisify(mkdir);
+const tempCleanup = Promise.promisify(track);
```
pimterry: You can promisify `track` if you like, but looking at the docs it doesn't take a callback or return anything at all, so I probably wouldn't :smile:.

tsconfig.json
hedss: I followed from [this](https://github.com/Microsoft/TypeScript/issues/12222) thread, as I was originally attempting to use `typeRoots`.

Adding it to the gulp file appears to cause Visual Code to complain as it doesn't find them. I'll use `includes` if that's better than `baseUrl`.

typings/temp.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'temp' {
+    function track(): void;
+    function mkdir(prefix: string): (err: Error, dirPath: string) => void;
```
pimterry: `mkdir` *returns* a callback? Seems unlikely - this should be the 2nd argument, and mkdir probably returns void.

This suggests the type you're getting when you `promisify` this isn't right either atm. Looks like the only reason you don't hit an implicit any is because the only place you call it you then immediately specify the next promise argument type.

Promisify does return the correct promisified type if the function you pass to promisify takes a callback as the last argument, but if it doesn't then it'll return a plain `Function` (i.e. callable, but takes and returns `any`). The types are little hacky but kind of interesting, if you're into that sort of thing: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/bluebird/index.d.ts#L390-L403. It means it works most of the time, but it's not totally foolproof. Implicit any should catch any substantial issues generally though.

tslint.json
hedss: We did; I'm also going to be dumping this file once I've got `resin-tslint` done, so tbh, I'm not hugely bothered. Shall change.

lib/versionbot.ts
hedss: That's a good point.

tslint.json
CameronDiver: Ah ok - makes sense.

Issue #7, opened 2017-02-06
Versionbot should comment on a PR should its status check fail 


hedss:
It should specifically ping the committer to let them know that a relevant `Change-Type:` tag should be included.

[hq edit] connects to https://github.com/resin-io/hq/pull/443

hedss:
In and working.

Issue #3, opened 2017-01-25
VersionBot: Could be used to notify PR originators that a build has failed


hedss:
@pimterry made the suggestion in resin-io/resin-ui#28, but essentially when a status check fails,  we could automatically comment on the status pinging the originator of the PR so they know.

lurch:
...and if this is added, a user might want the ability to 'unsubscribe' from these build-failure notifications? (maybe, I'm just guessing)

Also, there might need to be a way to configure this per status-check-service, as an (entirely different) repo on which I'm a contributor uses Codecov which automatically comments on the PR itself.

lurch:
And just to be pedantic, the original suggestion was in https://github.com/resin-io/resin-ui/pull/324  ;-)

hedss:
@lurch There's the ability to turn off VersionBot in a PR (`procbots/versionbot/no-checks` label when the PR is opened). Apart from that, a single comment is now provided on a PR the first time incorrect commits are seen (when the PR is opened). The status check will stay in a failed state but no further comments will be issued.

By commenting as soon as the PR is opened, it should be obvious in a notification list that something needs correcting. I think this is valid, as VersionBot never spams the author.

lurch:
Hmmm, looking at things from the other point of view (sorry!) if the user **is** relying on VersionBot to notify them of failing status-checks, they *might* expect this behaviour:  (and I'm assuming here that each status-check takes a significant amount of time to run, i.e. the user goes off and does something else in the meantime)
 * User submits their PR
   * status-check fails, user gets a notification message
 * User updates their PR
   * status-check fails again, and according to your last message they don't get a second notification
 * As the user didn't get a status-check failed notification, they assume it succeeded (when in fact it didn't)

?

Then again, perhaps I'm confusing/overlapping what the status-check service should do, with what VersionBot should do.

hedss:
Status checks for commits/merges in VersionBot are quick, less than a couple of seconds after the Github hook has fired. Jenkins is the real slowpoke, and you're going to come back for that anyway.

Almost as soon as a PR is opened, VersionBot will let you know if a commit is incorrect, which is why it only comments directly at the author once. This means the next commit should include a correct commit type. If it doesn't, the author's forgetful. The fact that the status itself will still fail and stop you from merging a PR is the ultimate reminder, just as Jenkins failures is a 'this code is broken'.

Because of this, I don't see any advantage to flagging a failing `versionist` commit more than once in the PR.

[...]

PR #8, opened 2017-02-06
Add comment on failed version status check, going to the raiser of the event that started the check

hedss:
Connects to #7

Change-Type: minor


hedss:
@pimterry Absolutely! I'm keeping #3 open for that reason.

I really need to get Versionbot up and running this week, so I'm going to do a few 'hard and fast' changes to do this, then when it's doing its job, I can come back to these more generic issues and solve them properly!

Issue #9, opened 2017-02-06
Procbots command labels should follow consistent format


hedss:
We wish to use labeling going forward to control Procbot implementations. This should follow the:

`procbots/<bot>/<command>`

format.

[hq edit] connects to https://github.com/resin-io/hq/pull/443

lekkas:
Just adding a note here on updating the README as well to reflect the changes to the commands

lekkas:
Also, given that merging is handled by `versionbot`, shouldn't 

`triggerLabels: [ 'flow/ready-to-merge' ]` 

be changed to `triggerLabels: [ 'procbots/versionbot/ready-to-merge' ]` as well?

(https://github.com/resin-io-modules/resin-procbots/blob/master/lib/versionbot.ts#L95

PR #10, opened 2017-02-06
Update command label syntax to consistent format.

hedss:
Connects to #9

Change-Type: patch


lib/versionbot.ts
```diff
@@ -85,20 +85,20 @@ export class VersionBot extends GithubBot.GithubBot {
...
                 workerMethod: this.checkVersioning
             },
             {
                 events: [ 'pull_request' ],
                 name: 'CheckVersionistCommitStatus',
-                suppressionLabels: [ 'flow/no-version-checks' ],
-                triggerLabels: [ 'version/check-commits' ],
+                suppressionLabels: [ 'procbots/versionbot/no-checks' ],
+                triggerLabels: [ 'procbots/versionbot/check-commits' ],
                 workerMethod: this.checkVersioning
```
pimterry: Independent of this change, but I don't understand why we have two `CheckVersionistCommitStatus` blocks here, which look identical except the second requires triggering. What does the second do that the first doesn't already cover?

hedss: Yeah, this is a hang-up from when I changed around where labels/events were filtered. Good catch, the second is now completely obsolete.

PR #14, opened 2017-02-07
Fix reauthentication issue and tagging problem.

hedss:
* Promise resolve is called on successful completion of operation and not post-catch block (stops `undefined` being passed to original caller).
* Tags of more than one digit are now usable for each semver section(!)

Connects to #13

Change-type: minor


lib/versionbot.ts
```diff
@@ -389,7 +389,7 @@ export class VersionBot extends GithubBot.GithubBot {
             // Now we get the new version from the CHANGELOG (*not* the package.json, it may not exist).
             return exec(`cat ${versionData.fullPath}${_.last(moddedFiles)}`).then((contents: string) => {
                 // Only interested in the first match for '## v...'
-                const match = contents.match(/^## (v[0-9]\.[0-9]\.[0-9]).+$/m);
+                const match = contents.match(/^## (v[0-9]+\.[0-9]+\.[0-9]+).+$/m);
```
lekkas: `versionist` currently doesn't have an option to return the calculated semantic version based on the commit history right?

lib/githubbot.ts
```diff
@@ -251,7 +252,7 @@ export class GithubBot extends ProcBot.ProcBot<string> {
                         // If there's more retries, try again in 5 seconds.
                         setTimeout(runApi, 5000);
                     }
-                }).then(resolve);
+                });
             };
```
pimterry: Maybe this is overkill, but this block could be simpler with some TS's async/await + a little bluebird goodness. Something like:

```typescript
protected gitCall = async (method: any, options: any, retries?: number): Promise<any> => {
  let badCreds = false;
  let retriesLeft = retries || 3;

  while (true) {
    retriesLeft -= 1;
    try {
      return await method(options);
    } catch (err) {
      if ((err.message === 'Bad credentials') && !badCreds) {
        badCreds = true;
        await this.authenticate();
      } else if (retriesLeft === 0) {
        throw err;
      } else {
        await Promise.delay(5000);
      }
    }
  }  
}
```

Pretty sure this gives the same behaviour (including getting this fix automatically, since you don't have to `resolve` by hand), but imo reads much easier.

pimterry: Oooh, in fact that refactor has just shown me a bug: if you have `retriesLeft === 1` at the start of `runApi`/the while loop, and you then get your first authentication error, then you skip the `retriesLeft === 0` check, end up with `retriesLeft === -1`, and this function retries forever without terminating.

Either way, we should definitely fix that (probably by checking `retries === 0` before the auth failure test, not after)

lib/versionbot.ts
hedss: Not that I know of, no.

lib/githubbot.ts
hedss: Good catch! There was actually a lot wrong with the method in general, and I've altered it, which will be in next push (in a sec!). I've not moved to async yet, I've never used it in anger and I'd like to understand it properly before using it. I'm going to add this as an issue which I can correct once the panic of this week is over, if that's ok? :)

lib/versionbot.ts
lekkas: Ok cool, it'd be great if versionist could help here so that, in case we change `versionist.conf.js`, we don't have to remember changing the fixed version parsing logic here as well. (Just a note, I don't think we should block here for the first version, so lgtm)

lib/githubbot.ts
pimterry: Sure, yeah, that bug definitely needed fixing (:+1:), but even long-term making this clear and readable is the important bit, async/await or otherwise.

I'll take this moment to plug this great intro to async/await if you want more background: https://www.youtube.com/watch?v=Hn-gkM0ARZU :smile:

Or Jake Archibald has a great guide (as ever) which might be a more efficient route: https://developers.google.com/web/fundamentals/getting-started/primers/async-functions

lib/versionbot.ts
hedss: Yep, def. sounds good to me!

Issue #18, opened 2017-02-09
VersionBot should delete the branch after a successful merge


hedss:
Currently doesn't.

Issue #12, opened 2017-02-07
Procbots should be able to notify Flowdock inboxes


hedss:
For both success and errors.
This is temporarily going to be a straight use of the FD API, but will graduate to using Hubot eventually as the 'one true comms source'.

PR #20, opened 2017-02-09
Add a FlowdockAdapter mixin; delete branch post-merge.

hedss:
This has also forced a restructure of the repo layout to make things clearer:
* Bots are now in their own dedicated directory
* There is now a 'mixins' directory, as I anticipate more for different roles
* There's now a `runtime.ts` which currently contains the approved way of creating classes based on mixins
* Sneaked label changes in at the same time
* Ensure on a successful merge, VersionBot posts to Flowdock (if enabled) as a non-error.
* Update the README
* Add a safety check for Flowdock validation
* Branch for PR is now deleted automatically once a PR is merged.
* Update README.md

Connects to #9
Connects to #12
Connects to #18

Change-type: minor


hedss:
I've been a bit naughty, I've sneaked in three fixes into one PR. However, I've done this out of an attempt to make this the last review before we start using this earnestly. Apologies.

In my defence, two of the issues have very small changes attached.

lib/mixins/flowdock.ts
```diff
@@ -0,0 +1,21 @@
...
+            json: true,
+            url: `https://api.flowdock.com/messages/team_inbox/${item.roomId}`,
+        };
+
+        // If we aren't outputting to Flowdock (development build, test build, etc)
+        // just return.
+        // People invariably use different cases, so...
+        if (process.env.FLOWDOCK_ALERTS && (process.env.FLOWDOCK_ALERTS.toLowerCase() === 'true')) {
+            // We don't wait for a response, we have no control over it anyway.
+            request.post(requestOpts);
```
pimterry: Might be nice to console.log errors though, so we can spot & debug any errors if they do turn up.

lib/mixins/flowdock.ts
```diff
@@ -0,0 +1,21 @@
+import * as request from 'request-promise';
+import { FlowdockInboxItem } from './flowdock-types';
+
+export class FlowdockAdapter {
+    public postToInbox(item: FlowdockInboxItem) {
```
pimterry: Why use this as a mixin, rather than a standalone object that VersionBot uses?

You could have `this.flowdock = new FlowdockAdapter()` in the VersionBot constructor, call `this.flowdock.postToInbox()` to get at it, get the same effect with less magic. Feels like a good case for composition over inheritance.

lib/mixins/flowdock.ts
hedss: We do, if you look at both the GithubBot and VersionBot actions, all logging still occurs. It's just that in the `flowdock` mixin it gets dropped. See `reportError()` in `versionbot.ts` as an example of this.

lib/bots/githubbot.ts
```diff
@@ -166,7 +166,8 @@ export class GithubBot extends ProcBot.ProcBot<string> {
                     return action.workerMethod(<GithubAction>action, data);
                 }).catch((err: Error) => {
                     // We log the error, so that it's saved and matches up with any Alert.
-                    this.alert(ProcBot.AlertLevel.ERROR, `Error thrown in main event/label filter loop: ${err.message}`);
+                    this.alert(ProcBot.AlertLevel.ERROR, 'Error thrown in main event/label filter loop:' +
+                        `${err.message}`);
```
hedss: Yep, there was more originally, forgotten to remove.

lib/mixins/flowdock.ts
hedss: The main reason is that currently it's just FlowDock, and in the future I'm assuming we'll support this, email, possible Front, and other external services perhaps.

So it was more futureproofing, it's easier now, but initialising a load of different services like this in the future instead of just adding the particular mixin as a new `implements` seems cleaner. If you feel strongly that we shouldn't be doing this, I'm happy to remove it. :)

README.md
```diff
@@ -79,40 +79,48 @@ Download the key and then create a Base64 string from it. It will be required la
...
     `WEBHOOK_SECRET`: The 20 digit hex key used to authenticate messages.
     `INTEGRATION_ID`: The ID given on Integration creation, a unique identifier.
     `PROCBOTS_PEM`: The Base64 encoded private key generated on Integration creation.
+	`FLOWDOCK_ALERTS`: If present and set to 'true', alerts will be sent to any relevant Flowdock room.
+
+`VersionBot` requires the following environment variables:
+
 	`VERSIONBOT_NAME`: The name shown in commits and merges for PRs by the Integration.
 	`VERSIONBOT_EMAIL`: Email address for the bot, (can be an empty string).
+	`VERSIONBOT_FLOWDOCK_ROOM`: The room ID whose inbox will be posted to. If `FLOWDOCK_ALERTS` is set to `true` this **must** be present.
```
hedss: Yeah, I'd ummed and ahhed about this; I've been thinking about something else that would affect this, but in retrospect it's probably a bad idea.

lib/mixins/flowdock.ts
pimterry: For this post request though, we don't do anything I think? It'll never reach `reportError`, because we don't return the promise, and we don't `catch` it here either, so any failure to talk to flowdock disappears into the ether. Or am I missing some other connection between this and `reportError`?

hedss: Ah, you mean if the post doesn't return, not for the error condition triggering it. Sorry, misunderstood. Sure.

lib/mixins/flowdock.ts
pimterry: One field with `public X = new X()` per service feels simpler than `implements X` + `applyMixin` + the stub function field, per service, especially at runtime. I've seen mixins used when you want to pull in a set of functions to add to your external api, and that's a great use case, but I don't think they works well if you only need to use them internally like this.

I am all for future-proofing and a declarative approach ofc. If do we want to really flexibly add lots of other services though, you want a way to set up the config for each service up front (in a perfect world, with one shared config), and then have standardized postError/postSuccess functions you can call for all of them at once, so the `reportError` code is simpler and notifier-agnostic. An instance-based example:

```typescript
class VersionBot {
  constructor(notifierConstructors: NotifierConstructor[]) {
    // Might not be practical to do all config upfront like this, but if not you can
    // always do per-service setup outside and pass in notifier instances.
    this.notifiers = notifierConstructors.map((Notifier) => new Notifier({
      tags: [ 'devops' ],
      from_address: 'versionbot@resin.io',
      ...
    }));
  }

  reportError(error: Error) {
    this.notifiers.forEach((notifier) => notifier.postError(error))
  }
}

export = new VersionBot([
  FlowdockNotifier,
  EmailNotifier,
  ... 
]);
```

With mixins you can't do this. Each mixin adds at least one separate function to the class, and you can't loop over them, so you can't decouple yourself from the functions involved, and you can't do any setup before the postToInbox/postError/whatever call either, so you have to configure them individually at notify time, unless you bring in MixinFactories (ick).

Not saying we should jump to the above until this is a real problem, just that mixins won't scale as a nice solution for this kind of thing. Simple objects stay way more flexible.

hedss: Makes sense; have altered to have it as a class member in VersionBot instead.

lib/bots/versionbot.ts
```diff
@@ -331,8 +377,18 @@ export class VersionBot extends GithubBot.GithubBot {
             }).then(() => {
                 this.log(ProcBot.LogLevel.INFO, `Upped version of ${repoFullName} to ${newVersion}; ` +
                 `tagged and pushed.`);
+            }).catch((err: Error) => {
+                // Call the VersionBot error specific method.
+                this.reportError({
+                    brief: `Versionbot failed to create a new version ready for merge for ${repoFullName}#${pr.number}`,
```
lekkas: nit: I find this message a bit hard to understand, can we rephrase these messages to something simpler like

`Versionbot failed to merge ${repoFullName}#${pr.number}`

After all we don't know which operation failed

Issue #32, opened 2017-02-15
Github Integration access for private repos requires auth token as password and username


hedss:
Without this, cloning of a repository will fail.

Issue #31, opened 2017-02-15
VersionBot: Github API prevents PR review comments from being retrieved on a private repo


hedss:
This prevents us from a safety check before merging, meaning that we have to rely on maintainers only adding a label once a review is carried out.

A question has been lodged [here](https://platform.github.community/t/cannot-retrieve-pr-reviews-from-a-private-repository/1346).

hedss:
Unfortunately this will be staying open for a while. Github have now responded, agreed it doesn't work, and removed the PR review comment API from both public **and** private repos.

hedss:
However! There may well be a workaround by looking at the issue timeline and matching up request for changes and approvals chronologically by user.

hedss:
![](https://github.trello.services/images/mini-trello-icon.png) [VersionBot: Github API prevents PR review comments from being retrieved on a private repo](https://trello.com/c/LWL6BYNV/10-versionbot-github-api-prevents-pr-review-comments-from-being-retrieved-on-a-private-repo)

hedss:
Refresh to use issue timeline to allow review check. Independent work, but related to #27

hedss:
This has been implemented properly in Github, and reviews are now accessible on both public and private repos.

PR #33, opened 2017-02-15
Patch for broken private Github API use and repo cloning

hedss:
Connects to #31 
Connects to #32


hedss:
Test note: These fixes have been tested on the live VersionBot Integration via the use of both a private and public test repo.

lib/bots/versionbot.ts
```diff
@@ -422,7 +428,7 @@ export class VersionBot extends GithubBot.GithubBot {
         };
 
         return Promise.mapSeries([
-            `git clone https://${process.env.WEBHOOK_SECRET}:x-oauth-basic@github.com/${versionData.repoFullName} ` +
+            `git clone https://${this.authToken}:${this.authToken}@github.com/${versionData.repoFullName} ` +
```
CameronDiver: I assume you've tested, but should this be repeated here?

lib/bots/githubbot.ts
```diff
@@ -61,7 +62,7 @@ export class GithubBot extends ProcBot.ProcBot<string> {
         // The `github` module is a bit behind the preview API. We may have to override
         // some of the methods here (PR review comments for a start).
         this.githubApi = new GithubApi({
-            // debug: true,
+            //debug: true,
```
CameronDiver: Nit: Comment should have a space after `//`, perhaps this one shouldn't be there at all though?

lib/bots/versionbot.ts
hedss: Yes, this is the reason, in fact, private repos didn't clone. See #32

Issue #34, opened 2017-02-15
VersionBot should specifically check for outlying status before running `versionist`


hedss:
As seen in resin-io/resin-image-maker#46, if Jenkins fails to complete then a manual merge/tag step is required.

Issue #19, opened 2017-02-09
VersionBot should always check after a new status update, in case it needs to merge


hedss:
When a new status check occurs, VersionBot should also carry out a check to see if this frees it for merge.

Example case:
1. PR is opened, check occurs, Versionist check is fine
2. Jenkins starts carrying out a build
3. PR is approved and label is added
4. Jenkins build fails

It could be that this is not the fault of the change, but a Jenkins stuffup.

So, whilst the merge will fail because of the Jenkins status, VersionBot will not automatically try and merge again should the status change. We need to also check for merge conditions when a status change occurs.

PR #35, opened 2017-02-15
Check for status changes for all linked Integration services to ensure that all applicable statuses are valid before versioning up/merging

hedss:
This is quite a complex checkin. To ensure that VersionBot interacts correctly with
other Github Integrations (notably Jenkins), VersionBot will now:

* Refuse to attempt a version up if any protected branch statuses have not succeeded
* Refuse to merge, even with a valid label, should any protected branch statuses have not succeeded
* Listen to status events to ensure that, should all other conditions have been met (version upped, label applied),
  a merge only occurs when protected branch statuses have succeeded

This means that we now do not get into a state where a merge can fail and require manual intervention.

Also adds a datestamp for logging/alerting output and tides up some old code.

Connects to #19
Connects to #34

Change-type: minor


lib/bots/versionbot.ts
```diff
@@ -654,6 +713,124 @@ export class VersionBot extends GithubBot.GithubBot {
...
+                        // We go ahead and merge.
+                        return this.mergeToMaster({
+                            commitVersion: commitMessage,
+                            owner,
+                            prNumber: prInfo.number,
+                            repoName: repo
+                        }).then(() => {
+                            // No tags here, just mention it in Flowdock so its searchable.
+                            // It's not an error so doesn't need logging.
+                            const flowdockMessage = {
```
lurch: Super-small nitpick, but I guess there's no need to construct the `flowdockMessage` if `process.env.VERSIONBOT_FLOWDOCK_ADAPTER` isn't set?

lib/bots/versionbot.ts
```diff
@@ -654,6 +713,124 @@ export class VersionBot extends GithubBot.GithubBot {
...
+                    if (commitMessage) {
+                        // We go ahead and merge.
+                        return this.mergeToMaster({
+                            commitVersion: commitMessage,
+                            owner,
+                            prNumber: prInfo.number,
+                            repoName: repo
+                        }).then(() => {
+                            // No tags here, just mention it in Flowdock so its searchable.
+                            // It's not an error so doesn't need logging.
```
lurch: ...and yet there's a `this.log` just a few lines further down??

Issue #37, opened 2017-02-16
VersionBot should ignore the unlabeling of a PR


hedss:
Currently it doesn't, so it checks to try and do a merge for both labeling (ie. `procbots/versionbot/ready-to-merge`) and *any* unlabeling. It should ignore the latter, as nothing VersionBot does is affected by it, and if it does see the merge label is set, it will attempt to carry out a merge again (this won't corrupt anything, but will flag an error).

We see this as some users have scripts running labeling/unlabeling PRs as their status changes.

lurch:
What about the situation where the `procbots/versionbot/no-checks` label gets removed - the user might then expect VersionBot to do the checks it'd been previously told to skip?

lurch:
> We see this as some users have scripts running labeling/unlabeling PRs as their status changes.

Maybe this is another situation where we need to insert a time-gap? i.e. if a label gets removed, wait for e.g. a couple of minutes to see if it gets added back again (in which case no action should be taken), before responding to the unlabelling event (if the label didn't get re-added)? IYSWIM.

Apologies if you've already considered that and I'm just adding 'noise'.

hedss:
Should the label get removed then a check will occur the next time a status update occurs or a synchronization is forced (and in fact if the `procbots/versionbot/ready-to-merge` label is applied).

Re-labelling events are automatic, they will be seen so we do not need to wait for them.

Closing this issue as it's working as it should.

lurch:
Perhaps it'd be worth adding something to the docs to explain how to 'manually force' VersionBot to run checks after removing the `procbots/versionbot/no-checks` label?
(or perhaps I'm still just misunderstanding things!)

hedss:
There is a section in the PR and commit guidelines spec. that mentions how to kick VersionBot to restart.

Issue #6, opened 2017-02-06
Remind PR commenters should a PR 'languish' for some time


hedss:
Sometimes we get into a situation where a PR that's currently open can languish.

It would be good to have a Procbot that goes through open PRs and notes the last time any activity occurred. It could then specifically raise a comment that mentions all reviewers and the originator.

Obviously this is only useful on PRs that have the `flow/review` label attached, all others should still be ignored.

This Procbot (maybe Monitorbot) would need to have timer functionality, rather than be driven purely by Github hooks.

lurch:
...and if it's passed the minimum number of reviews, and none of the maintainers have added the `procbots/versionbot/ready-to-merge` label yet, then maybe it could only ping the maintainers and not the reviewers? ;-)

hedss:
Yep, that's a good idea!

PR #38, opened 2017-02-16
Ensure that unlabeling of PRs is ignored.

hedss:
This ensures that checks to see if a PR is mergable is not carried out.
Whilst VersionBot won't attempt to create a new merge, it will complain that
it can't. This means that scripted labeling/unlabeling of a PR will have an
error attached at the end of a successful merge. We need to quash this.

Also lower number of retries to a total of 3.

Connects to #37

Change-Type: patch


lurch:
> will have an error attached at the end of a successful merge

Is there any way for VersionBot to check if a PR has already been merged?

hedss:
No, the closest you can get is look for an unattached head. However, this is actually a pointless check, as VersionBot shouldn't even be triggered in this case (which this issues fixes).

lurch:
Related question: Is VersionBot able to detect (and ignore) closed PRs? Or when a PR gets closed without merging, does the user need to remember to add the `procbots/versionbot/no-checks` label?

hedss:
No, it only acts on particular PRs which are being opened, synchronized or labeled. It ignores everything else.

To be clear, VersionBot has a very specific role, the versioning up and merging of PRs. There may well be (and we are discussing) other ProcBot based bots that will have other tasks.

lurch:
The reason I ask, is because if #6 gets implemented, it might get annoying if VersionBot (and/or some other ProcBot) starts sending automated reminders about PRs which were closed without merging.
(but again, perhaps that's something you've already thought of)

hedss:
#6 would not be implemented by VersionBot. There'd be a StatusBot or the like that would get information on all the repos it currently has attached to it and then has a timeout to go and check all **open** PRs (not PRs that have been closed without/with merging).

Again, this comes down to selectiveness about what a ProcBot listens to. There's a registration scheme in the ProcBots that means any derived Bots **only** get what they want, they don't get all events. This is how we can tailor this. Maybe it's worth reading the ProcBot spec and the source code?

Issue #40, opened 2017-02-16
Inconsistent labels in README


lurch:
`README.md` currently mentions both the `procbots/versionbot/ready-for-merge` label and the `procbots/versionbot/ready-to-merge` label - I'm assuming one of these is a typo?

lurch:
Fixed by #41

Issue #39, opened 2017-02-16
VersionBot should only check for a `Change-Type` tag on the footer of a commit.


hedss:
Without doing this, `versionist` will fail.

lurch:
Is this something that should be checked by `versionist` itself, rather than VersionBot? \*shrug\*

hedss:
@lurch I and @lekkas have discussed this, and ideally it absolutely should. I can't find the relevant issue right now, but yes. What should be happening is `VersionBot` pulls in `versionist` as an NPM module and uses it programatically, getting any errors/failures back and then using them to percolate up the chain.

Unfortunately, up to now `versionist` has been a bit of a moving target, so executing it as command seemed the easiest way to ensure compatibility.

PR #41, opened 2017-02-17
Ensure that only the footer for a commit message includes the `Change-Type:` tag; ensures Flowdock posting on successful merges.

hedss:
* VersionBot erroneously looked for the `Change-Type:` tag anywhere in the commit message. This meant
  that it verified commits even when `versionist` could not find them. This no longer occurs.
* VersionBot now correctly posts to the relevant Flowdock inbox should the envvar be set when a
  merge successfully occurs.
* Update the README.md

Connects to #39
Connects to #40

Change-Type: patch


lib/bots/versionbot.ts
```diff
@@ -212,11 +212,38 @@ export class VersionBot extends GithubBot.GithubBot {
...
+                if (index > 0) {
+                    // We should have a line index to join from, now.
+                    lines.splice(0, index);
+                    const footer = lines.join('\n');
+                    const invalidCommit = !footer.match(/^change-type:\s*(patch|minor|major)\s*$/mi);
+
+                    if (!invalidCommit) {
+                        changetypeFound = true;
+                        break;
+                    }
```
CameronDiver: If we push through https://github.com/resin-io/versionist/issues/62 it would mean that we could use that to check that the commit was valid, keeping versionist and VersionBot tightly coupled.

hedss: In fact, if you look at resin-io-modules/resin-procbots#39, we're *all* on the same track here. :)

CameronDiver: Haha nice.

lekkas: I agree with @CameronDiver that versionist/VersionBot should be more tightly coupled but this needs more than just pushing through resin-io/versionist#62 ; `versionist` should be properly refactored so that it can be integrated as a module rather than an executable command (@hedss and I earlier during this week). The reason is that issues like this will keep coming up and now we are forced to duplicate our efforts (like in this particular case). 

That said, this is the ProcessBot PoC week so we can let this slip through for now before fixing the versionist/ProcessBots integration and start paying off technical debt next week

lib/bots/versionbot.ts
```diff
@@ -212,11 +212,38 @@ export class VersionBot extends GithubBot.GithubBot {
...
+                //  <footer>
+                // This code will actually let you get away with:
+                //  <header>
+                //
+                //  <footer>
+                // As sometimes a development patch may be self-explanatory in the header alone.
+                const lines = commitMessage.split('\n');
+                let index = 0;
+                for (index = lines.length - 1; index >= 0; index -= 1) {
+                    if (lines[index].match(/$\s*^/)) {
```
pimterry: This is interesting. Given this is true, this PR won't work at all - `index` will always be 0 after this loop, so we'll never find a change-type in any commit, and this PR will entirely stop the bot in its tracks.

...Can I use this to nudge us towards unit & long-term maybe even integration testing for this? With core infrastructure like this, it'd be nice if it was really hard to totally break it without realising..

hedss: A good catch.. yet it's worked with all the tests I've thrown at it.
Yes, unit and integration tests are on the list, but it was deemed more important to get them going first. Lots of things on the list...

pimterry: > A good catch.. yet it's worked with all the tests I've thrown at it. Which seems odd.

Ah no, got it! It _is_ possible for `/$\s*^/` to match, but only against the empty string, because then you're matching against `^` and `$` at the same time in the same place, with no spaces. This regex can still work, but only as long as it's on an actually empty line. It's just the whitespace-allowing part that will never match. Neat.

Issue #46, opened 2017-02-21
GithubBot `gitCall()` method should check error type as it may not be an object


hedss:

hedss:
Currently a non-repository based event will cause an error. We should just use a generic context for the `ProcBot` at this point.

PR #47, opened 2017-02-21
Ensure that non-repo based events are not treated as an error; fix typo so merge messages report the bot correctly.

hedss:
Also adds a missing semi-colon typo.

Connects to #46
Connects to #43

Change-Type: patch


lib/bots/versionbot.ts
```diff
@@ -226,7 +226,7 @@ export class VersionBot extends GithubBot.GithubBot {
                 //  <footer>
                 // As sometimes a development patch may be self-explanatory in the header alone.
                 const lines = commitMessage.split('\n');
-                const lastLine = _.findLastIndex(lines, (line) => { return line.match(/^\s*$/) });
+                const lastLine = _.findLastIndex(lines, (line) => { return line.match(/^\s*$/); });
```
pimterry: Nit: This could use the arrow function expression syntax to drop the brackets/`return`/etc:

```javascript
const lastLine = _.findLastIndex(lines, (line) => line.match(/^\s*$/));
```

Issue #51, opened 2017-02-23
ProcBot method returns aren't explicitly typed


hedss:
And should be.

Issue #50, opened 2017-02-22
Versionbot doesn't ignore already merged PRs


pimterry:
See https://github.com/resin-io/resin-ui/pull/450#discussion_r102524450, where a comment on the PR after merge triggered an error.

hedss:
This is down to the `procbots/versionbot/ready-to-merge` label still being assigned, and because a `pull_request` event is being fired on comments. Essentially, the label should be automatically removed on a successful merge.

hedss:
It does now, the unlabeling occurs and further comments do not provoke it.

lurch:
Do not provoke the VersionBot, lest you feel its wrath.

hedss:
More of a nasty suck, tbh. :)

lurch:
Do not provoke the VersionBot, lest it spams multiple identical comments on your PRs :laughing:

Issue #45, opened 2017-02-21
ProcBots should use a configuration file, if available


hedss:
The configuration file is written in YAML and includes nesting for each hierarchy of bot.
For example, the top level class is the `ProcBot`, which is currently inherited by `GithubBot` (which knows how to deal with the Github API and events as an Integration) and which itself is inherited by `VersionBot`.
Therefore we can create a YAML file that is processed by the `ProcBot` class, but hold information for each of the children:
```
procbot:
    minimum_version: 0.3.2
    githubbot:
        versionbot:
            minimum_approvals: 2
            approved_reviewers:
                - lekkas
                - hedss
                - Page-
                - petrosagg
                - pimterry
            maintainers:
                - lekkas
                - hedss
```
There is a 'retrieveConfiguration()' method at the top level (which in `ProcBot` looks for a file path), that retrieves the configuration and then calls `processConfiguration()` which returns the processed config. `processConfiguration()` is itself not designed to be overriden and converts the YAML into an object.

In the flow for a dummy ProcBot child that does nothing, then, it would call `retrieveConfiguration()` which would return any suitable object.
In the flow for a `GithubBot` derived child, it calls `retrieveConfiguration()` which is overridden by `GithubBot` and takes a repo owner, name and branch. It then retrieves the configuration from this repo and calls it's parent `ProcBot` with `processConfiguration()` before returning it.

At each level, if a class understands that a particular property relates to a boolean (for example, `minimum_version`), it can determine the result of this and inject this as the value for the field in the returned object. This allows conditional tests to be performed on the configuration object to allow children to determine whether they should run or not.

In the above example then, `minimum_version` would be determined as `true` if the `ProcBot` running were `0.3.2` or above, or `false` otherwise.

lurch:
> In the above example then, `minimum_version` would be determined as `true` if the `ProcBot` running were `0.3.2` or above, or `false` otherwise.

Hmm, but what if a ProcBot actually _needs_ the original `minimum_version` value? What if the 'booleanized result' was instead stored in `minimum_version_result` ? \*shrug\*

Does `processConfiguration()` do any 'flattening' of the YAML hierarchy, or is the nesting preserved? i.e. can VersionBot see the `minimum_version`, or is that accessible only by ProcBot?

If a YAML file stored configuration entries for multiple bots, e.g.
```yaml
procbot:
    minimum_version: 0.3.2
    githubbot:
        versionbot:
            minimum_approvals: 2
[...]

PR #52, opened 2017-02-23
Add VersionBot configuration files via Github repos; ensure `procbots/versionbot/ready-to-merge` label is removed post-merge

hedss:
Github repos may now explicitly control how VersionBot behaves, via the use of
a YAML file. Currently, a full, valid file consists of:

```
procbot:
    minimum_version: <semver number>
    githubbot:
        versionbot:
	    maintainers:
	        - <Github username 1>
		- <Github username 2>
		- ...
```

All contents are optional.
* Should a `minimum_version` be found, ProcBot children will not run should their
  version not at least match the value given.
* Should a list of maintainers be given, VersionBot will not attempt to carry
  out a merge should the `procbots/versionbot/ready-to-merge` label not have been
  added by a user in the `maintainers` list.

All code methods also now explicitly define a return type.

Connects to #45
Connects to #51
Connects to #50

Change-type: minor


hedss:
This is the last set of features and fixes before a full test-suite is added. It's really got too big now, and I'm getting seriously worried about regressions. These fixes should satisfy us for ongoing VB functionality for a while.

lurch:
Could you add code-blocks to your first comment so that indentation is preserved? Ta.

README.md
```diff
@@ -119,7 +119,24 @@ Ensure you also create the `procbots/versionbot/ready-to-merge` label type in th
 
 Finally you need to install the Integration into the repo. Do this by going to your 'Settings' page, selecting 'Installed Integrations', selecting your Integration and then selecting the repos you want it installed in in the 'Repository access' section.
 
-## Running
+### Tailoring ProcBots for a Repo via Configuration File
+
+ProcBots now also respond to a configuration file. This is a file with the name `.procbots.yml` in a relative location for the ProcBot running. In the case of the VersionBot, this is in the root of the repository that it is working on.
+
+The configuration file uses a set of nested properties based on the class hierarch of the ProcBots, with each class able to modify variables at run time from the configuration file, eg:
```
lurch: hierarch -> hierarchy

README.md
```diff
@@ -119,7 +119,24 @@ Ensure you also create the `procbots/versionbot/ready-to-merge` label type in th
 
 Finally you need to install the Integration into the repo. Do this by going to your 'Settings' page, selecting 'Installed Integrations', selecting your Integration and then selecting the repos you want it installed in in the 'Repository access' section.
 
-## Running
+### Tailoring ProcBots for a Repo via Configuration File
+
+ProcBots now also respond to a configuration file. This is a file with the name `.procbots.yml` in a relative location for the ProcBot running. In the case of the VersionBot, this is in the root of the repository that it is working on.
```
lurch: Should this specify "In the case of the VersionBot, this is in the root of the repository that it is working on, on the `master` branch." ?  (just for removal of any doubt)

lurch: ...and is this just for VersionBot, or is it the default behaviour for anything derived from GithubBot?

README.md
```diff
@@ -119,7 +119,24 @@ Ensure you also create the `procbots/versionbot/ready-to-merge` label type in th
...
+ProcBots itself has a single property, `minimum_version`, which is checked to ensure that operations only get carried out should that version be satisfied.
+
+Example:
+
+```
+procbot:
+    minimum_version: 0.5
+```
+
+Other bots are free to add the use of any properties they require. It should be noted that all derived bots are able to view the entire configuration file on request.
```
lurch: Remove "the use of" ? (i.e. "Other bots are free to add any properties they require.")

README.md
```diff
@@ -129,15 +146,30 @@ This allows the checking of commits for a PR and merging them when the right lab
...
 
-Current, `VersionBot` will:
+Currently, `VersionBot` will:
 * Version up only once all status checks set as 'Required' on protected branches for 'master' are successful
-* The `procbots/versionbot/ready-to-merge` is present
+* The `procbots/versionbot/ready-to-merge` label is present
+
+VersionBot can be configured via `.procbots.yml` file present in any repo that it operates on. This alters the settings for it for working on that repository. Currently the configuration for VersionBot consists of the following properties:
+
+    `procbot.githubbot.versionbot.maintainers` - A list of Github user names that are authorised maintainers of the repository. The `procbots/versionbot/ready-to-merge` label will only be acted upon should a maintainer in this list have added the label.
```
lurch: Just out of curiosity, what happens if `procbot.githubbot.versionbot.maintainers` gets added but is (accidentally) left empty - does that mean VersionBot will never be able to merge any PRs?

README.md
```diff
@@ -119,7 +119,24 @@ Ensure you also create the `procbots/versionbot/ready-to-merge` label type in th
...
+
+The configuration file uses a set of nested properties based on the class hierarch of the ProcBots, with each class able to modify variables at run time from the configuration file, eg:
+
+ProcBots itself has a single property, `minimum_version`, which is checked to ensure that operations only get carried out should that version be satisfied.
+
+Example:
+
+```
+procbot:
+    minimum_version: 0.5
```
lurch: Did the auto-conversion to boolean get added? If so, should that be documented in the README?

README.md
```diff
@@ -129,15 +146,30 @@ This allows the checking of commits for a PR and merging them when the right lab
 
 `VersionBot` will ignore any status checks and not attempt to merge should the `procbots/versionbot/no-checks` label be present on any PR it would otherwise operate on when a PR is opened.
 
-Current, `VersionBot` will:
+Currently, `VersionBot` will:
 * Version up only once all status checks set as 'Required' on protected branches for 'master' are successful
-* The `procbots/versionbot/ready-to-merge` is present
+* The `procbots/versionbot/ready-to-merge` label is present
+
+VersionBot can be configured via `.procbots.yml` file present in any repo that it operates on. This alters the settings for it for working on that repository. Currently the configuration for VersionBot consists of the following properties:
```
lurch: "via a `.procbots.yml` file" ?

lib/bots/githubbot-types.d.ts
```diff
@@ -38,3 +39,9 @@ export interface GithubActionRegister extends GithubAction {
 
 // A GithubActionMethod is the method that will be used to process an event.
 export type GithubActionMethod = <T>(action: GithubAction, data: T) => Promise<void>;
+
+export interface GithubBotConfiguration extends ProcBotConfiguration {
+    githubbot: {
+        versionbot?: any;
```
pimterry: Similar to my below comment about `ProcBotConfiguration`, this should be the minimum interface we want to depend on at this stage, and adding `any` properties defeats that. I'd drop this for now.

lib/bots/procbot.ts
```diff
@@ -113,6 +119,36 @@ export class ProcBot<T> {
...
+            return;
+        }
+
+        // Swap out known tags that become booleans.
+        const minimumVersion = ((config || {}).procbot || {}).minimum_version;
+        if (minimumVersion && process.env.npm_package_version) {
+            config.procbot.minimum_version = (process.env.npm_package_version < minimumVersion) ? false : true;
+        } else {
+            config.procbot.minimum_version = true;
+        }
```
pimterry: I find it pretty strange that minimum version change from a number to a boolean. If you immediately fail here, rather than delegating using the boolean result to `.getConfiguration` later on, I don't think you'd need to do this - it's always a number, and you either move forwards, or give up immediately.

lib/bots/procbot.ts
```diff
@@ -113,6 +119,36 @@ export class ProcBot<T> {
...
+
+        return config;
+    }
+
+    // Retrieve a configuration file.
+    // This default implementation assumes a pathname.
+    protected retrieveConfiguration(path: string): Promise<ProcBotConfiguration> | Promise<void> {
+        return fsReadFile(path,).call('toString').then((contents) => {
```
pimterry: Nit: Extra comma in `fsReadFile(path,)`

lib/bots/procbot.ts
```diff
@@ -113,6 +119,36 @@ export class ProcBot<T> {
...
+            return this.processConfiguration(contents);
+        }).catch(() => {
+            this.log(this._logLevel.INFO, 'No config file was found');
+            return;
```
pimterry: Nit: this return doesn't do anything, could equivalently drop it.

lib/bots/versionbot.ts
```diff
@@ -84,12 +84,24 @@ interface FSError {
...
+}
+
 // The VersionBot is built on top of GithubBot, which does all the heavy lifting and scheduling.
 // It is designed to check for valid `versionist` commit semantics and alter (or merge) a PR
 // accordingly.
 export class VersionBot extends GithubBot.GithubBot {
     // Flowdock Adapter
     private flowdock: FlowdockAdapter;
+    private mergeLabel = 'procbots/versionbot/ready-to-merge';
+    private ignoreLabel = 'procbots/versionbot/no-checks';
```
pimterry: Should these be constants?

lib/bots/versionbot.ts
```diff
@@ -435,6 +454,13 @@ export class VersionBot extends GithubBot.GithubBot {
                 throw new Error(`alreadyCommitted`);
             }
 
+            // If this was a labeling action and there's a config, check to see if there's a maintainers
+            // list and ensure the labeler was on it.
+            if ((data.action === 'labeled') && botConfig) {
+                // Note that labeling can only occur on a PRE data, hence casting.
```
pimterry: You can use [discriminated unions](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) to explain this to the compiler and then it'll do the casting automatically in cases like this, and give you extra safety in some others too.

lib/bots/versionbot.ts
```diff
@@ -825,12 +861,23 @@ export class VersionBot extends GithubBot.GithubBot {
...
-                            prNumber: prInfo.number,
-                            repoName: repo
+                        // Ensure that the labeler was authorised. We do this here, else we could
+                        // end up spamming the PR with errors.
+                        return this.getConfiguration(owner, repo).then((config: VersionBotConfiguration) => {
+                            // If this was a labeling action and there's a config, check to see if there's a maintainers
+                            // list and ensure the labeler was on it.
+                            // This throws an error if not.
+                            if ((data.action === 'labeled') && config) {
+                                this.validMaintainer(config, data);
```
pimterry: You don't need the `&& config` check, both here and before.

`validMaintainer` already handles the case where there's no config or the config doesn't have maintainers. Its arg types don't show it right now, but if you just make VersionBotConfig optional you can drop this.

lib/bots/versionbot.ts
```diff
@@ -845,15 +892,56 @@ export class VersionBot extends GithubBot.GithubBot {
...
+                            this.log(ProcBot.LogLevel.INFO, `MergePR: Merged ${owner}/${repo}#${prInfo.number}`);
                         });
                     }
                 });
             }
         });
     }
 
-    private reportError(error: VersionBotError) {
+    private validMaintainer(config: VersionBotConfiguration, event: GithubBotApiTypes.PullRequestEvent): void {
```
pimterry: nit: this method doesn't have a verb in the name. `ensureEventIsFromMaintainer`? `verifyEventSender`?

lib/bots/versionbot.ts
```diff
@@ -845,15 +892,56 @@ export class VersionBot extends GithubBot.GithubBot {
...
+            // If there is a config file and a minimum version, ensure this can run.
+            if (config) {
+                // If there's no minimum version, then we're fine.
+                const minimumVersion = (config.procbot || {}).minimum_version || true;
+                if (!minimumVersion) {
+                    throw new Error('${process.env.VERSIONBOT_NAME} cannot run as its version is less than the ' +
+                        'required minimum for the ${owner}/${repo} repository. ' +
+                        `This: ${process.env.npm_version_number}, Req: ${minimumVersion}`);
+                }
+            }
```
pimterry: Why doesn't the initial processConfiguration do this? It lives at the top-level in the config, and it feels like it should be checked at the top-level in the code, rather than per-bot.

README.md
```diff
@@ -129,15 +146,30 @@ This allows the checking of commits for a PR and merging them when the right lab
...
+
+    `procbot.githubbot.versionbot.maintainers` - A list of Github user names that are authorised maintainers of the repository. The `procbots/versionbot/ready-to-merge` label will only be acted upon should a maintainer in this list have added the label.
+
+Example:
+
+```
+procbot:
+    minimum_version: 0.5
+    githubbot:
+        versionbot:
```
lurch: I think the implication might be that in future there might also be `githubbot`-specific configuration?

lib/bots/procbot.ts
lurch: Yeah, that's also something I'd asked about in #45

lib/bots/versionbot.ts
lurch: Similarly, perhaps `versionBotCommits` should be renamed too?

lib/bots/procbot.ts
hedss: I'd considered that maybe a child bot could override in extenuating circumstances if there's an absolute need to do something. But maybe it does make more sense to just fail the processing and throw an error.
Will change.

lib/bots/versionbot.ts
hedss: Yep.

lib/bots/versionbot.ts
hedss: WRT the previous comment about minimum version, this should be removed and an error thrown in the processing.

README.md
hedss: Yes, I'd quite like to keep this for exactly that reason, especially if there's a test in the `githubbot` block that causes it to fail for all children. It's immediately obvious then when adding/modifying properties.

lib/bots/versionbot.ts
pimterry: Doesn't need to live in the class though. It's not something you want exposed, and it's not mutable state, so it could just be constants that live above. Let's you drop this `this.` everywhere too.

This is bikeshedding territory though. `readonly` or `const` is fine by me.

lib/bots/versionbot.ts
hedss: That is very neat. Like it.

lib/bots/versionbot.ts
hedss: JavaScript I got into, fine. TypeScript is still occasionally causing me headaches. I'm still trying to break out of the C++ mentality, so up the top it'd have turned into a static (or worse a global), hence keeping it in the class.
I'll get there eventually, I promise!

hedss: I think you're right, and not just for `this`. It also means all predefined `const`s for the class are at the top of source, so it's easy to find them. I'll move them out.

README.md
hedss: It'll be counted as a non-existent list and won't be used; anyone will be able to start a merge.

README.md
hedss: It didn't, I've quashed it. :)

README.md
```diff
@@ -121,11 +121,11 @@ Finally you need to install the Integration into the repo. Do this by going to y
...
+ProcBots now also respond to a configuration file. This is a file with the name `.procbots.yml` in a relative location for the ProcBot running. In the case of the GithubBot (and any other child bot derived from it, such as VersionBot), this is in the root of the repository that it is working on, on the `master` branch.
 
-The configuration file uses a set of nested properties based on the class hierarch of the ProcBots, with each class able to modify variables at run time from the configuration file, eg:
+The configuration file uses a set of nested properties based on the class hierarchy of the ProcBots, with each class able to modify variables at run time from the configuration file.
 
-ProcBots itself has a single property, `minimum_version`, which is checked to ensure that operations only get carried out should that version be satisfied.
+ProcBots itself has a single property, `minimum_version`, which is checked to ensure that operations only get carried out should that version be satisfied. Should it not find itself to be of at least the minimum version, and error is thrown.
```
lurch: "and error" -> "an error"

lurch: Hmm, and I wonder if it'd be worth extending this to the "child classes" too, so that GithubBot and VersionBot in `.procbots.yml` could each specify their `minimum_version`? (although obviously at this early stage it's maybe a bit overkill)

lib/bots/procbot.ts
```diff
@@ -130,9 +130,9 @@ export class ProcBot<T> {
         // Swap out known tags that become booleans.
         const minimumVersion = ((config || {}).procbot || {}).minimum_version;
         if (minimumVersion && process.env.npm_package_version) {
-            config.procbot.minimum_version = (process.env.npm_package_version < minimumVersion) ? false : true;
-        } else {
-            config.procbot.minimum_version = true;
+            if (process.env.npm_package_version < minimumVersion) {
+                throw new Error('Current ProcBot implementation does not minimum required version for the operations');
```
lurch: "does not meet minimum" ?

lurch: (and I wonder if this error message should specifically mention `.procbots.yml` rather than a vague "the operations"?)

lib/bots/versionbot.ts
```diff
@@ -822,7 +828,7 @@ export class VersionBot extends GithubBot.GithubBot {
     }
 
     // Has VersionBot already made commits to the branch.
-    private versionBotCommits(prInfo: GithubBotApiTypes.PullRequest): Promise<string | null> {
+    private hasVersionBotCommits(prInfo: GithubBotApiTypes.PullRequest): Promise<string | null> {
```
lurch: Just to be pedantic, I'd expect a method named `has...` to return a boolean, rather than a string?

lib/bots/versionbot.ts
```diff
@@ -900,16 +906,16 @@ export class VersionBot extends GithubBot.GithubBot {
         });
     }
 
-    private validMaintainer(config: VersionBotConfiguration, event: GithubBotApiTypes.PullRequestEvent): void {
+    private isValidMaintainer(config: VersionBotConfiguration, event: GithubBotApiTypes.PullRequestEvent): void {
```
lurch: And again to be pedantic, I'd aslo expect a `is...` method to return a boolean, so perhaps `checkValidMaintianer` would be a better name here?

lib/bots/versionbot.ts
```diff
@@ -84,24 +85,29 @@ interface FSError {
...
     };
 }
 
-// The VersionBot is built on top of GithubBot, which does all the heavy lifting and scheduling.
+type GenericPullRequestEvent = GithubBotApiTypes.PullRequestEvent | GithubBotApiTypes.PullRequestReviewEvent;
+
+const MergeLabel = 'procbots/versionbot/ready-to-merge';
+const IgnoreLabel = 'procbots/versionbot/no-checks';
+
+// Te VersionBot is built on top of GithubBot, which does all the heavy lifting and scheduling.
```
lurch: Te -> The

lib/bots/procbot.ts
hedss: Not convinced; if a config's been set up and is running, then the owners should know that the config is where it is.

README.md
hedss: Possibly in future, but for now I think a single place is enough.

lib/bots/procbot.ts
lurch: Yeah, I hadn't considered that the first person to see this error would likely be the person who had most recently edited `.procbots.yml` ;-)

lurch: ...but I still think you ought to change the message from:
`'Current ProcBot implementation does not minimum required version for the operations'`
to
`'Current ProcBot implementation does not meet minimum required version for the operations'`

(i.e. insert the missing `meet` word)

lib/bots/procbot-types.d.ts
```diff
@@ -0,0 +1,11 @@
+// ProcBot --------------------------------------------------------------------
+
+// ProcBot configuration format.
+// This includes any ProcBot class specific variables, as well as any
+// potential child bot configurations (of type `any`, ProcBot does not care
+// what these may be)
+export interface ProcBotConfiguration {
+    procbot: {
+        minimum_version?: number | boolean;
```
pimterry: We can now drop the `| boolean` part of this I think.

lib/bots/versionbot.ts
```diff
@@ -425,16 +450,23 @@ export class VersionBot extends GithubBot.GithubBot {
...
+            return this.getVersionBotCommits(prInfo);
         }).then((commitMessage: string | null) => {
             if (commitMessage) {
                 throw new Error(`alreadyCommitted`);
             }
 
+            // If this was a labeling action and there's a config, check to see if there's a maintainers
+            // list and ensure the labeler was on it.
+            if ((data.action === 'labeled') && (data.type === 'pull_request')) {
+                // Note that labeling can only occur on a PRE data, hence casting.
```
pimterry: This comment can go now, since there's now no casting.

lib/bots/versionbot.ts
```diff
@@ -425,16 +450,23 @@ export class VersionBot extends GithubBot.GithubBot {
...
             // Ensure we've not already committed. If we have, we don't wish to do so again.
-            return this.versionBotCommits(prInfo);
+            return this.getVersionBotCommits(prInfo);
         }).then((commitMessage: string | null) => {
             if (commitMessage) {
                 throw new Error(`alreadyCommitted`);
             }
 
+            // If this was a labeling action and there's a config, check to see if there's a maintainers
+            // list and ensure the labeler was on it.
```
pimterry: This comment talks about "and there's a config" but you're no longer doing that check here.

Issue #56, opened 2017-03-01
Add circle integration


brownjohnf:

lekkas:
Can we close this @brownjohnf ?

brownjohnf:
Yep, sorry. I failed to reference it correctly.

PR #57, opened 2017-03-01
[Issue #56] Add travis integration

brownjohnf:


Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+	&& rm -rf /tmp/*
+
+# We only copy the actual code, none of the development or credential files.
+# Everything else we need is from Envvars.
+COPY bin ./bin
+COPY build ./build
+
+# server.js will run when container starts up on the device
+ENTRYPOINT ["/usr/local/bin/node", "./bin/procbots"]
+CMD ["-b", "versionbot"]
```
brownjohnf: This will allow us to deploy the same image with different run commands to run the different bots, as we add them.

.travis.yml
```diff
@@ -0,0 +1,16 @@
+---
+sudo: required
+git:
+  depth: 1
+services:
+  - docker
+pre_install:
+  - docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
```
brownjohnf: This build will fail, because I haven't figured out how to get the docker creds from apo

brownjohnf:
Okay, I couldn't get travis to work. I switched to circle and it's a dream. I'm going to open a new PR.

Issue #62, opened 2017-03-08
VersionBot: Multi-status Integrations can cause status check to fail


hedss:
See: https://github.com/resin-io/node-lkl/pull/17

It appears that the branch protection has a single Travis entry, but the number of Travis status checks can differ. This breaks the check algorithm, causing VersionBot to fail.

lurch:
From a quick rummage, looks like https://github.com/resin-io-modules/resin-procbots/blob/master/lib/bots/versionbot.ts#L819 is the troublesome line? (but of course I could easily be wrong)

hedss:
Yes it is, but I need to check to ensure multi-status would be fixed by always being '>=', which given the Travis layout, I'm not sure.

lurch:
Yeah, I guess if you had integrations from Travis, AppVeyor and Jenkins, then VersionBot would (naively?) be looking for at least 3 status checks. But if Travis and AppVeyor returned two status checks each, then VersionBot might think it's "good to merge" even if the Jenkins status-check hasn't completed yet?
Obviously you could change the logic to check that each integration has returned at least one status-check, but that might still go wrong if the first Travis check passes, and VersionBot thinks things are good-to-merge before the second Travis check fails.

Perhaps this is something else that needs to be configurable via `.procbots.yml` ?  (if you can't get the info you need from the Github API)

hedss:
You keep reading my mind. :)

The main problem here is, again, Github's API has changed since they first designed it, which has lead to backwards compat. lockin. I'm going to see if there's another way I can get min. potential checks somehow.

lurch:
...and as I think I might have mentioned elsewhere, there's also some services (e.g. CodeCov) where you want the status-checks displayed for informational purposes, but you might not want a failing CodeCov status check to actually block the VersionBot merging process. (but obviously that's a bridge that's probably not worth crossing until we actually get to it)

hedss:
In fact for things such as the latter, it's not an issue. If a status is not included in the protected branch settings, then it's not flagged by Github and VB won't look for it.

The other thing I should have mentioned was that statuses are all cleared when a new commit occurs, so we won't get into a situation where VB okays something if a check hasn't been carried out. You can actually see this in how it currently works, where after it commits a new version bump, it has to wait until Jenkins has finished building.

lurch:
> In fact for things such as the latter, it's not an issue. If a status is not included in the protected branch settings, then it's not flagged by Github and VB won't look for it.

[...]

PR #63, opened 2017-03-13
Ensure that multi-status services used on protection branches are correctly handled

hedss:
Previously status entries for PRs were not correctly interrogated to ensure that
all statuses were met. Single CI services can, in fact, set multiple status fields
on a PR, so all need to be checked.

Also adds a specific check for PR validity before attempting to merge. This is
required because should status checks complete after a merge (for those not in
protected mode) then VersionBot can whinge.

Connects to #62

Change-Type: patch


hedss:
Note that this is vital for those services (such as Travis) where they offer multiple status entries but only have a single protected entry.

lib/bots/versionbot.ts
```diff
@@ -802,22 +808,44 @@ export class VersionBot extends GithubBot.GithubBot {
...
+            // required status contexts (which we do anyway), then go through each
+            // actual status check from the combined, and try and match the prefix of a context
+            // with each status. If we get a hit, and the status is a failure, then we
+            // have failed. If we match and the status is a pass, we've passed.
+            // We can therefore assume that a pass has occurred if:
+            //  * We have seen one of every context in the protected status list at least once
+            //    AND
+            //  * Each of those seen has passed
+            // Should any protected context not be seen in the current status checks, then
+            // we have failed.
```
lurch: So, if `travis-ci/push` has a valid status-check (but `travis-ci/pr` is still building), and then VersionBot runs, VersionBot will still think everything is 'okay' because `travis-ci` has at least one good status-check, even if `travis-ci/pr` eventually comes back with a failing status-check?
(I've never looked at the GH API, but I wonder if there might be some kind of 'pending' status-check value? Which would correspond to the 'orange circle' that the GH web UI displays)

Also, how would this work on the [Etcher repo](https://github.com/resin-io/etcher/pulls) where we have TravisCI running separate builds for multiple arches (Linux x86, Linux x64 and OSX x64 - see [an example](https://travis-ci.org/resin-io/etcher/builds/210704103) ? Or do they just get coalesced into a single pass/fail from the GH point of view?

lib/bots/versionbot.ts
```diff
@@ -802,22 +808,44 @@ export class VersionBot extends GithubBot.GithubBot {
...
-                }
+                });
             });
-            if ((foundStatuses.length !== contexts.length) ||
-                _.includes(foundStatuses, false)) {
+
+            // Have a final list of results, if any of them have a failure state, we
+            // bail and log it so we can see what went wrong.
+            if (!_.every(statusResults, [ 'passed', true ])) {
+                this.log(ProcBot.LogLevel.WARN, `Status checks failed: ${JSON.stringify(statusResults)}`);
```
lurch: I wonder if it'd be useful to store more info in `statusResults`, rather than only storing the boolean result of `status.state === 'success'` ?

Also, if e.g. Travis passes, but Appveyor fails, would it be worth filtering out the passing-checks and only displaying the failing ones?

lurch:
And one final question - if a status-check fails because of a fault of the CI service (e.g. a timeout on Travis) and so the status-check returns false, VB will obviously then fail. But if the author of the PR then restarts the CI build, and this time it passes, does the author have any way of triggering VB that it should re-run its checks (given that no new commit was needed), or do they need to wait for a maintainer to add the appropriate label? (apologies if I'm misunderstanding the full VB flow)

hedss:
The maintainer (or author as is more commonly the case) re-applies the label. What should actually happen is the CI service realises it failed.

lurch:
> The maintainer (or author as is more commonly the case) re-applies the label.

Sorry, I should have clarified that I was referring to the situation (forgive me for forgetting the correct terminology) where VersionBot is setup to ignore labels that aren't applied by a maintainer (and in my example, the author isn't a maintainer). Again, having not actively used versionist and VersionBot myself, there's a good chance I'm misunderstanding things...?
(In a nutshell, I guess what I'm asking is: is there any way for a non-maintainer author to get VB to re-run its checks, without having to explicitly push a new commit to the PR?)

> What should actually happen is the CI service realises it failed.

That's not always possible. For example on Etcher, we have a couple of Appveyor build processes, and as part of the appveyor-build-script it needs to install some dependencies. But when there's some network timeout error, Appveyor just returns "build failed" (I guess it just checks the exit-code?) and doesn't know _why_ the build failed. So the PR shows a failing status check, but when you click through to the actual appveyor log, and see it failed due to a network timeout, manually restarting the appveyor build often works the second time, and when the re-build finishes, the github PR then shows a passing status-check.

hedss:
> (In a nutshell, I guess what I'm asking is: is there any way for a non-maintainer author to get VB to re-run its checks, without having to explicitly push a new commit to the PR?

No. And there shouldn't be. It is absolutely not the remit of VB to work round a failing third-party.

There should be a way of doing this from a Github comment. Jenkins, for example, has the 'test this please' command, which kicks a build manually. I'd be surprised if the same wasn't true for Appveyor. Travis, for example, allows you to restart a build from it's Dashboard (or from the CLI).

Worst comes to the worst, create an empty commit.

lurch:
> Travis, for example, allows you to restart a build from it's Dashboard

Yes, Appveyor also has a "restart build" button on its web UI, that's exactly what I meant by "manually restarting the appveyor build", apologies for any confusion.

> Worst comes to the worst, create an empty commit.

Heh, fair enough.

P.S. Did you also see the inline-comments I left further up on `lib/bots/versionbot.ts` ?

lurch:
P.P.S. Just out of curiosity, what happened [here](https://github.com/resin-io/resin-ui/pull/558) ? Looks like VB complained that there was no Jenkins build-status, but then merged anyway, and there's no indication that anybody re-applied any labels? Does VB actually get notified (and re-triggered) when a status-check finishes running?

lib/bots/versionbot.ts
hedss: There is indeed a 'pending' check. At the point where a PR is opened Travis reports how many statuses are to be applied. So if one has passed but the other hasn't, then we see a 'pending' and VB marks it as failed.

If there are multiple builds that run on a PR, then they'll all get their own status. So it won't work until all pass.

lib/bots/versionbot.ts
hedss: Thought about that, but tbh, if something fails the only sensible way I'm going to be able to determine exactly why something's gone wrong is to list all the statuses so I know exactly what was expected to run. Otherwise a 'fail' on its own doesn't help.

hedss:
Just seen the other comments, have commented. :)

Strangely, it looks like something else to do with the GH API has either changed or we get a extra status check now. We do checks when a new status change occurs, so even though this is an annoying warning it doesn't affect actually merging the issue.

lib/bots/versionbot.ts
lurch: But, to be pedantic, you're not listing all the statuses, you're only listing the result of all the `status.state === 'success'` checks ;-)

lurch:
Hmmm, given that "There is indeed a 'pending' check" and "We do checks when a new status change occurs" would it be worth combining those and modifying VB so that it doesn't spam a PR with a "status check failed / missing" message until none of the status-checks are returning a `pending` status?

hedss:
Yes, that's the idea. I just haven't got to it yet. :(

Issue #68, opened 2017-03-16
ProcBots re-architecture required for multi-listener/emitter framework


hedss:
SyncBots will require a multi-listener/emitter framework. And it would improve from the class hierarchy we have at the moment.

PR #71, opened 2017-03-22
ProcBots: Move from hierarchy based bots to added service listeners/emitters

hedss:
ProcBots no longer inherit from a class hierarchy of the services they wish to use.
Instead, they now add listeners/emitters to listen for events/emit data to a service.

This is an important architectural change to allow bots to receive events from/dispatch messages to multiple services. This allows greater flexibility for bots and makes them truly 'general purpose'.

Listeners:
 * Listen on a chosen comms protocol for events. These are now queued similarly to before
   except that listeners are Workers, and not the bot itself.
 * Send events in order for a context
 * Send both the raw and cooked (if possible) data to a registered client via a ServiceEvent object

Emitters:
 * Take ServiceEmitRequest objects detailing contexts. Should a given context include
   data that allows an emitter to send that data on, then it does so
 * Return a ServiceEmitReponse object with any data returned (or errors)

ProcBots class:
 * Deals with the adding of listeners/emitters and also has methods to dispatch to emitters
 * Takes a ServiceEmitRequest for config files, if required, to use an emitter to retrieve a config file

Clients:
 * Register for the listeners/emitters that they require, using a similar registration process
   to before
 * Get ServiceEvent objects when events come in from a service
 * Send ServiceEmitRequest objects to the dispatcher for a named service (or all services) when required

Connects to #68 

Change-Type: major


hedss:
And as an aside, I am updating the ProcBots spec. to include this.

brownjohnf:
If I'm understanding this correctly, it appears that this is essentially a pub/sub broker, without support for asynchronous reads, history/persistence, and remotely registered clients. Is that correct?

hedss:
Well, in the sense that any event listener/message broker pattern is pubsub, yes. It's not a single pubsub channel set, however. You have to subscribe to the listeners you want, and they'll send you data, but you still have to know how to handle it, and how to send it on. It also handles contexts for you, in the case of the `github` service, for example, it knows how to queue events so they don't get out of order for a particular repo, but apart from that allow execution in any order.

There is no persistence. Persistence would occur, if required, by use directly by a client bot.

lib/bots/versionbot.ts
```diff
@@ -102,108 +103,166 @@ interface StatusResult {
...
+    protected statusChange = (registration: GithubRegistration, event: ServiceEvent): Promise<void | void[]> => {
         // We now use the data from the StatusEvent to mock up a PullRequestEvent with enough
         // data to carry out the checks.
-        const githubApi = this.githubApi;
-        const owner = data.name.split('/')[0];
-        const repo = data.name.split('/')[1];
-        const commitSha = data.sha;
-        const branches = data.branches;
+        const owner = event.cookedEvent.data.name.split('/')[0];
+        const repo = event.cookedEvent.data.name.split('/')[1];
```
CameronDiver: I'd consider using a regex, or just `split`ting once at the least.

lib/bots/versionbot.ts
```diff
@@ -269,70 +334,86 @@ export class VersionBot extends GithubBot.GithubBot {
...
                 // We only complain on opening. Either at that point it'll be fixed, or we
                 // simply don't pass it in future. This stops the PR being spammed on status changes.
-                if (data.action === 'opened') {
-                    this.gitCall(githubApi.issues.createComment, {
-                        body: `@${data.sender.login}, please ensure that at least one commit contains a` +
-                            '`Change-Type:` tag.',
-                        owner,
-                        number: pr.number,
-                        repo: name,
+                if (event.cookedEvent.data.action === 'opened') {
```
CameronDiver: Would it not be better to just warn the user of not having a change-type footer when it comes to merge time? My thinking is that PRs often have a lot of conversation going on, and it would be better to have the warnings (essentially PR metadata) occur after the main bulk of discussion, after approvals etc. Not a big sticking point for me but is probably worth thinking about.

lib/bots/versionbot.ts
```diff
@@ -269,70 +334,86 @@ export class VersionBot extends GithubBot.GithubBot {
...
+                    repo: name
+                },
+                method: ghApiCalls.issues.getIssueLabels
             });
-        }).then((labels: GithubBotApiTypes.IssueLabel[]) => {
+        }).then((labels: GithubApiTypes.IssueLabel[]) => {
             // If we don't have a relevant label for merging, we don't proceed.
             if (_.filter(labels, (label) => {
                 return label.name === MergeLabel;
             }).length !== 0) {
```
CameronDiver: I know it's not part of the PR but I'd use `_.any` for this, means that you don't have to loop through all if you don't need to.

tsconfig.json
```diff
@@ -15,7 +15,10 @@
     "include": [
         "typings/**/*.d.ts",
         "lib/**/*.ts",
+        "lib/apis/**/*.ts",
         "lib/bots/**/*.ts",
-        "lib/mixins/**/*.ts"
+        "lib/framework/**/*.ts",
+        "lib/services/**/*.ts",
+        "lib/utils/**/*.ts"
```
CameronDiver: You can just include `lib/` and `typings/` but I suppose there's no harm in being explicit.

lib/bots/versionbot.ts
hedss: I'd rather have a warning as soon as it opens than when I came to merge it if it's been approved and then told 'no'. For point of reference, VersionBot hasn't changed at all in operation since it's last PR. It's only the framework changes that have been required in the move that have altered it.

tsconfig.json
hedss: I did actually have just that before, but I personally prefer being explicit. I can see the pros of both, tbh.

lib/bots/versionbot.ts
hedss: Good point. Did you mean `_.some()` or is the lodash docs out of date?

CameronDiver: Sorry yes I did, monday morning...

docs/procbot-architecture.md
```diff
@@ -0,0 +1,322 @@
...
+
+Workers are self-regulating; every time an event has been processed by a Client Bot the Worker will check to see if there are anymore events to pass to a Client. It will do this until the queue is empty and then remove itself from the map of Workers held in the WorkerClient.
+
+ServiceListeners all rely on the WorkerClient class to perform the context duties for queueing new events.
+
+### ServiceEmitters
+
+ServiceEmitters act as a Client Bots ability to send data to external services. This may be a simple 'dumb' fire and forget, or the act of using an API to request data and receive responses.
+
+Every ServiceEmitter expects a `ServiceEmitRequest` object which holds data allowing it to communicate relevant to the external service it acts as an arbiter for. As a `ServiceEmitRequest` object could potentially be passed to many ServiceEmitters, they contain a `contexts` property, which should contain a unique property named by the service that the Client Bot wishes to use, the value of which is an object that allows a service to make requests to that service.
```
lekkas: > As a `ServiceEmitRequest` object could potentially be passed to many ServiceEmitters, they contain a `contexts` property,

'they contain' -> 'it' has, this refers to the 'ServiceEmitRequest' object and not to ServiceEmitters right?

> which should contain a unique property named by the service that the Client Bot wishes to use, the value of which is an object that allows a service to make requests to that service.

Does 'allows a service' refers to the ServiceEmitter service?

Question: can ServiceEmitter instances be reused among ClientBots?

lib/bots/versionbot.ts
hedss: NP, ta. :)

docs/procbot-architecture.md
hedss: Good spot. Yep, the `ServiceEmitRequest` has the `contexts` property. I need to rewrite that sentence to make sense!

As to the question, I had considered only instantiating one instance of a `Service[Listener/Emitter]` and then sharing it between all bots that want one. However, I decided against this and went for new instances for each Bot for a couple of reasons:
1. It makes each Bot self-contained. If, for some reason, a particular Service died because of something a bot listened for/emitted, it won't take down out any other data another Bot might also be passing through it at the time.
2. It makes handling of sub-service infrastructure (such as Express usage) a lot easier to handle.

Happy to discuss the pros and cons of this if there's strong opinion!
I'm happy to discuss

Issue #74, opened 2017-04-13
ProcBots: Worker queuing allows parallel execution of same context


hedss:
This should be serial execution to allow for race conditions.

PR #75, opened 2017-04-13
Ensure that the queue is only reduced after a WorkerMethod executes and that the next WorkerMethod executes in the same scope

hedss:
Connects to #74

Change-type: patch


hedss:
With thanks to @sqweelygig for finding the original issue!

PR #82, opened 2017-05-02
Ensure that `npm bin` is used to get the correct binary dependency path, allowing `versionist` to be called

hedss:
This fixes the issue under K8 which is causing failure when a PR should be version bumped.

Change-Type: patch


lib/bots/versionbot.ts
```diff
@@ -650,17 +650,21 @@ export class VersionBot extends ProcBot {
...
-            ], cliCommand);
+            let versionistCommand: string;
+            return this.getNodeBinPath().then((nodePath: string) => {
+                versionistCommand = `${nodePath}${path.sep}versionist`;
+                if (exists) {
+                    versionistCommand = `${versionistCommand} -c versionist.conf.js`;
+                    this.logger.log(LogLevel.INFO, 'Found an overriding versionist config ' +
+                        `for ${versionData.repoFullName}, using that`);
+                }
+                console.log(versionistCommand);
```
pimterry: Leftover console.log (I think? if we do want this, it should at least be `this.logger.log` instead)

lib/bots/versionbot.ts
```diff
@@ -650,17 +650,21 @@ export class VersionBot extends ProcBot {
...
-                    `for ${versionData.repoFullName}, using that`);
-            }
-
-            return Promise.mapSeries([
-                versionistCommand,
-                'git status -s'
-            ], cliCommand);
+            let versionistCommand: string;
+            return this.getNodeBinPath().then((nodePath: string) => {
+                versionistCommand = `${nodePath}${path.sep}versionist`;
```
pimterry: Nit: imo this is easier & clearer as just `path.join(nodePath, 'versionist')`. Also handles some cases like normalizing out any extra slashes (e.g. on the end of `nodePath`). Don't see how those could come up, but maybe that's useful :man_shrugging:

lib/bots/versionbot.ts
hedss: Yep, shouldn't be there. :)

lib/bots/versionbot.ts
hedss: Yep, better way to do it. :)

PR #86, opened 2017-05-08
Ensure that git clones are removed during merge attempts.

hedss:
Previously erroneous code meant that none of the cloned git repositories
that were created during a merge attempt were removed post-success/failure.
This fix ensures that they now are.

Change-type: patch


hedss:
Connects to #86

PR #88, opened 2017-05-09
Improve status checks for required external services; reduce verbose 'spamming' of PR on status failures.

hedss:
Previously, VersionBot was very verbose in its status checks; either everything succeeded
or there was a failure. This did not work well with pending statuses. More than that, it
made comments which weren't actually necessary as the Github status check dialogue makes
it obvious why a PR can't be merged.

Additionally, there are some services which actually don't even report a status such as
Jenkins which simply ignores PRs until it's polled and is ready for them. This has the
effect of Github not being able to return a status, or even an entry, for Jenkins as it
hasn't been acknowledged by it. We therefore still attempt to go through with a merge,
but don't report if the service is still outstanding. Jenkins eventually catches up, does
a build, reports a status and the PR is merged.

Connects to #62

Change-type: patch


hedss:
One for Tim and Lucian, here. ;-)

lib/bots/versionbot.ts
```diff
@@ -961,20 +966,25 @@ export class VersionBot extends ProcBot {
...
                         });
                     }
                 });
             });
 
-            // Have a final list of results, if any of them have a failure state, we
-            // bail and log it so we can see what went wrong.
-            if (!_.every(statusResults, [ 'passed', true ])) {
+            // If any of the checks are pending, we wait.
+            if (_.filter(statusResults, [ 'state', StatusChecks.Pending ]).length > 0) {
```
pimterry: Can just use `_.any`

lib/bots/versionbot.ts
```diff
@@ -961,20 +966,25 @@ export class VersionBot extends ProcBot {
...
+                return StatusChecks.Pending;
+            }
+
+            // If any of the checks didn't pass, we fail.
+            if (!_.every(statusResults, [ 'state', StatusChecks.Passed ])) {
```
lurch: Is this effectively the same as:
```js
if (_.any(statusResults, [ 'state', StatusChecks.Failed ])) {
```
?

lib/bots/versionbot.ts
```diff
@@ -538,13 +540,11 @@ export class VersionBot extends ProcBot {
...
-        }).then((statusesPassed) => {
+        }).then((checkStatus) => {
             // Finally we have an array of booleans. If any of them are false,
             // statuses aren't valid.
-            if (!statusesPassed) {
-                throw new Error(`At least one status check has failed; ${process.env.VERSIONBOT_NAME} will not ` +
-                    'proceed to update this PR unless forced by re-applying the ' +
-                    `\`${MergeLabel}\` label`);
+            if ((checkStatus === StatusChecks.Failed) || (checkStatus === StatusChecks.Pending)) {
+                throw new Error('checksPendingOrFailed');
```
lekkas: nit: should we consider using typed errors and de-multiplex based on the error type instead of the error message?

hedss: Yep, it's a better way of doing it. I've got that under some architectural changes I want to do in a big lump anyway. There's a few of these around, so I'd like to change them all at the same time.

PR #89, opened 2017-05-10
Flowdock Keep-alive

sqweelygig:
* Added keep-alive endpoint to Flowdock adapter
* Improved output of activated message listeners

Change-Type: patch


sqweelygig:
Better?

Issue #87, opened 2017-05-09
VersionBot: Status changes post-PR merge cause 'Pull request is not mergeable' message on PR


hedss:
There is a check to ensure that a status check coming in for a closed PR isn't merged, but for some reason this is failing. See: https://github.com/resin-io/procbots-private-test/pull/102 where the non-required Jenkins job kicked a status change post-merge.

PR #93, opened 2017-05-11
Ensure that multiple merge attempts fail cleanly.

hedss:
Enthusiastic merge label addition before a status check has completed
can potentially lead to two attempts to merge to occur (because both the
addition of the label and the status check passing occur within a small
timeframe of each other). The failing merge should fail silently as all
other checks to this point have shown that merging is valid.

Connects to #87

Change-type: patch


hedss:
This PR removes the highly annoying:
---
VersionBot failed to carry out a status check for the above pull request here: #111. The reason for this is:
Pull Request is not mergeable
Please carry out relevant changes or alert an appropriate admin.
---
message that can occur sometimes when label/status change checks overlap.

hedss:
If one of those occurred and the PR was still open, then there'd be a visible sign in the UI ('Branch is not up-to-date'/'Conflicts need resolving'/etc.).

However, I'll add an extra safety check for confidence purposes. :-)

hedss:
@pimterry And done!

Issue #94, opened 2017-05-11
ProcBots: Add ability to get external service handles directly from services


hedss:
The Github service currently passes the GH API instance around in events. This is not an ideal way to do things, as it implies that only listeners can give access to comms handles.

Either listeners or emitters should be able to do this, allowing the creation of suitable commands for those comms libraries (again, there is a GH API SDK we use for talking to GH, used by VersionBot).

PR #95, opened 2017-05-12
ProcBots: Add ability to retrieve underlying SDK API instance object/handle for a service

hedss:
Previously, should a service use an underlying SDK (such as a Node module)
to communicate externally with the service they provide access for, the
API instance would have to be communicated separately (in the case of VersionBot,
it used the API instance sent through cooked events). Whilst not used directly
for emitting/listening, there are instances where this is useful (for example to
form commands for an emitter, such as relevant method names and parameters).

This commit now allows Listeners/Emitters to expose the underlying SDK API instance
(should one exist) to allow bots to retrieve them more easily. It should be noted
that bots should *not* use the APIs directly to communicate with external services,
this should still occur via emitters.

Updates all current services/bots as required.

Connects to #94

Change-type: patch


lib/services/discourse.ts
```diff
@@ -199,6 +199,11 @@ export class DiscourseService extends MessageService implements ServiceListener,
     get serviceName(): string {
         return DiscourseService._serviceName;
     }
+
+    // There is no API handle for discourse.
```
sqweelygig: IMO new code should be TSDoc'd

lib/services/flowdock-types.d.ts
```diff
@@ -41,3 +39,8 @@ export interface FlowdockMessage {
     content: string;
     [key: string]: string;
 }
+
+// Handle type allowing Github API instance to be exported.
```
sqweelygig: This comment mentions Github, where it should be Flowdock.

lib/services/flowdock.ts
```diff
@@ -230,6 +231,13 @@ export class FlowdockService extends MessageService implements ServiceEmitter, S
     get serviceName(): string {
         return FlowdockService._serviceName;
     }
+
+    // Retrieve the API handle for Flowdock.
```
sqweelygig: New code should be TSDoc'd?

lib/services/github.ts
```diff
@@ -245,6 +244,13 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
         return this._serviceName;
     }
 
+    // Return the API handle.
```
sqweelygig: New code should be TSDoc'd?

lib/services/message-service.ts
```diff
@@ -189,4 +188,7 @@ implements ServiceListener, ServiceEmitter {
      * get the service name, as required by the framework
      */
     abstract get serviceName(): string
+
+    // Retrieve the API handle, if any.
```
sqweelygig: TSDoc?

lib/services/service-types.d.ts
```diff
@@ -64,16 +64,27 @@ export interface ServiceRegistration {
 // What a client of a listener implements when called. Returns any particular context.
 export type ServiceListenerMethod = (registration: ServiceRegistration, event: ServiceEvent) => Promise<void>;
 
+// The ServiceAPIIntances are a set of instances of service interfaces (if any).
+// These are used to allow, for example, the calling of specifically typed methods.
+// These are intended to be handles directly to SDK instances used to make requests
+// to/from a service. Comms going through interfaces such as HTTP(S) requests are not
+// intended to return a handle.
```
sqweelygig: TSDoc?

PR #97, opened 2017-05-12
Add TypeDoc documentation and build command to Resin ProcBots.

hedss:
This is an initial pass at documenting ProcBots with TypeDoc. It is by
no means exhaustive, but makes a good start on it.
The issue associated with this PR is intended to stay open until the
current extent of the project has been documented.

Connects to #96.

Change-type: patch


hedss:
As an aside, I believe this is our first attempt at using `TypeDoc`. If we're happy with the format, I'll add some info to the TypeScript Guide and add appropriate details to the skeleton project.

Note: I've not used the `gulp-typedoc` project, because it seems like it's lagging behind the current version, and threw errors which `tsc` and `tslint` didn't. Because of this, it currently uses `typedoc` directly.

gulpfile.js
```diff
@@ -37,5 +38,10 @@ gulp.task('tslint', () =>
         .pipe(tslint.report())
 );
 
-gulp.task('build', [ 'typescript', 'copydecs' ]);
+gulp.task('typedoc', () => {
+    exec('`npm bin`/typedoc --name "Resin ProcBots" --module commonjs --target ES6 --includeDeclarations ' +
+        '--excludeExternals --out docs/Typedoc lib/');
```
pimterry: Even if we include declarations, we should include `typings` as externals, since we _definitely_ don't want docs for those, and by changing the root they mess up all the file paths shown (making them lib/X and typings/Y, rather than just X). `--externalPattern '**/typings/*.d.ts'` seems to work (although it's strange and annoying to me that `**` is required there)

gulpfile.js
```diff
@@ -37,5 +38,10 @@ gulp.task('tslint', () =>
         .pipe(tslint.report())
 );
 
-gulp.task('build', [ 'typescript', 'copydecs' ]);
+gulp.task('typedoc', () => {
+    exec('`npm bin`/typedoc --name "Resin ProcBots" --module commonjs --target ES6 --includeDeclarations ' +
```
pimterry: I'm not sure we do want to include declarations, both generally (except in a few specific cases, mainly just if you have a type definition for something you expect downstream users to implement to use your module) or in this module specifically (we don't need the github type defs in our docs).

hedss:
@pimterry Good point about Pages, hadn't considered it. I've moved the other docs into appropriate resources directories and now reference them in the Typedoc generation command so they get included. Referenced from the main `README.md` which forms the front page.

Other points addressed as well.

PR #98, opened 2017-05-15
Added SyncBot getting started notes

sqweelygig:
A trivial one, but it would really help my adoption story to have this document to link to.

* Created early stage of SyncBot docs
* Tidied file name of example shell script

Change-Type: patch


sqweelygig:
@hedss Sorry for the unnecessary noise on this thread, it should now be done?

Issue #100, opened 2017-05-19
Github Service: Authentication doesn't occur on an emitter


hedss:
Whlist we create the API, we don't authenticate at that point, which means nothing's been tried by the time a handle user comes to use it.

PR #101, opened 2017-05-22
Ensure that authentication occurs on both Github ServiceListener and ServiceEmitter

hedss:
Also fixes issue which would cause VersionBot to fail.

Connects to #100

Change-type: patch


brownjohnf:
All the docs changes make reviewing this a mess. Will that happen on every PR from now on?

hedss:
@brownjohnf See #103

Issue #102, opened 2017-05-22
ProcBots: Typedoc changes large numbers of documents every change


hedss:
This is because the commit blob for sources changes. We don't actually need the sources referenced, so embedding a default theme without these would solve this.

PR #103, opened 2017-05-22
Add modified Typedoc command to link to sources from `master`

hedss:
Currently whenever we commit new code, the entire `docs` directory gets
regenerated with links to new commit blobs, even if the actual documentation
hasn't changed. This becomes extremely noisy.

Whilst having links to the source is nice, it's arguable that it's not strictly
required. And in fact, it would be lovely if Typedoc fixed this in another way,
though I have no idea how atm.

Connects to #102

Change-type: patch


hedss:
I've added a load of you on this, apologies, but it's to provoke a discussion.

Typedoc is something I'd like to adopt is possible (and add to the TypeScript Guide), but there are obviously issues like these we need to discuss to get there.

pimterry:
I do quite like the source links. One way we could keep them would be to use the `--gitRevision <?>` param (see https://github.com/TypeStrong/TypeDoc#theming) to fix the links to a specific branch, so they don't change with each commit. We could use `master`, and accept that that means docs for all older commits may end up with links to the wrong lines, but the current head of `master` will always be fine, and we'll never get a per-commit diff. Thoughts?

hedss:
@pimterry That's actually exactly what we want. I have no idea how I missed the option! That solves it, will redo this PR to use it.

hedss:
@pimterry Actually, this is a bigger issue, as sources will always lag behind by one PR. And that could be very confusing when linking to the sources. So, I'm not so convinced any more...

PR #111, opened 2017-05-31
Docs: Syncbot account (WIP)

imrehg:
Link to where Syncbot is found

### DIFF
--- a/docresources/SyncBot/syncbot-readme.md
+++ b/docresources/SyncBot/syncbot-readme.md
@@ -28,7 +28,7 @@ A little Bot, using the ProcBot framework, that links communication services
     * Become a Discourse Admin
     * Log out and in
 * Give your discourse details to SyncBot
-  * Send the following PM's to the SyncBot account:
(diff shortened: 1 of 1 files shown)


imrehg:
This change is not necessarily to be included here (since it's a public repo), but there should be some docs where to find Syncbot. That's the correct info, right, btw? (when I tried setting things up, it took a while to find).
Or if better, can change this patch to a "How to find Syncbot in your Flowdock setup" instead of the hot linking.

hedss:
@imrehg @sqweelygig @lekkas I'm of the opinion that anything that is general user information for how to use the service should be in `resin-io/process` somewhere (much like how to use VersionBot is in the Commit & PR guide).

lekkas:
agreed @hedss , `resin-io/process` is the place to add internal documentation on how to use syncbot cc @imrehg  @sqweelygig , so this issue should be closed and all relevant documentation transferred there

sqweelygig:
Added this task to the more general https://github.com/resin-io-modules/resin-procbots/issues/115

PR #112, opened 2017-05-31
Improve `ProcBot:dispatchToEmitter()` method to simplify calling/return conventions

hedss:
Previously a Client Bot had to construct a relevant ServiceEmitRequest before
calling `dispatchToEmitter()`, and then had to check the ServiceEmitResponse
on its fulfilment. This is overkill considering that a specifically named
ServiceEmitter is being invoked. Now, `dispatchToEmitter()` takes any relevant
data structure for the named emitter and, on response from the service, either
throws any returned error or fulfils with the data that was returned. This
simplifies the calling in a client, which may now just do:
```typescript
this.dispatchToEmitter(<serviceEmitterName>, <serviceDataStructure>)
```
Change-Type: major


hedss:
@sqweelygig Andrew, could you confirm that the changes I've made are good for SyncBot operation please? I think the only thing needed an alteration was `syncbot.ts` itself.

Then when your PR for Front goes in and is merged, we'll review this properly.

sqweelygig:
I am irritatingly close to getting my SyncBot PR in, can I please leave this review alone for a few working hours.

hedss:
There was no urgency to this. It was just to make sure before I put it in properly for review, there wasn't anything glaring from your side.

sqweelygig:
@hedss I've just taken a look and it looks good from my point of view.  We'll review it properly in a few days.

lib/bots/versionbot.ts
```diff
@@ -26,10 +26,9 @@ import { cleanup, track } from 'temp';
 import * as GithubApiTypes from '../apis/githubapi-types';
 import { ProcBot } from '../framework/procbot';
 import { ProcBotConfiguration } from '../framework/procbot-types';
-import { FlowdockEmitRequestContext } from '../services/flowdock-types';
-import { GithubCookedData, GithubEmitRequestContext, GithubHandle,
-    GithubRegistration } from '../services/github-types';
-import { ServiceEmitRequest, ServiceEmitResponse, ServiceEvent } from '../services/service-types';
+import { GithubError } from '../services/github';
```
sqweelygig: Is there a reason why GithubError doesn't live alongside the other github-types?

hedss: Yes, it's a class implementation, which can't exist in ambient declarations. It therefore has to be defined in a TS source file and not definition file.

lib/bots/versionbot.ts
```diff
@@ -354,7 +341,7 @@ export class VersionBot extends ProcBot {
         generateWaffleReference(pr.body);
 
         // Now look through all the commits in the PR. Do the same thing.
-        return this.githubCall({
+        return this.dispatchToEmitter(this.githubEmitterName, {
```
lurch: Possibly naive question, but rather than change every instance of `this.githubCall(...)` to `this.dispatchToEmitter(this.githubEmitterName, ...)`, would it have made sense to just change the `githubCall` method to something like
```
githubCall(args) {
    return this.dispatchToEmitter(this.githubEmitterName, args);
}
```
? (using whatever the relevant TypeScript syntax is)

hedss: Possibly, but the point was to remove an extraneous call that didn't need to be there. *shrug*

lurch: You're the maintainer - doesn't bother me either way ;)

hedss: Pfft. I'm sticking with what we go, then! :)

Issue #121, opened 2017-06-02
ProcBots: Documentation links don't exist in GH pages as prefixes are `_`


hedss:
See here:
https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/

Adding a `.nojekyll` file in the root of the repo should solve this.

PR #122, opened 2017-06-02
Add a `.nojeykll` file to ensure Github Pages includes TypeDoc module files

hedss:
Change-type: patch


hedss:
@imrehg Good catch, I'd put it in the wrong place. :)

lurch:
Was this missing a `connects-to:` ?

hedss:
It was indeed.

Connects-To: #121

Issue #99, opened 2017-05-16
VersionBot: Warn author should any of the status checks fail


hedss:
@flesler has suggested that an author of a PR should be pinged should a status check on a PR fail (author of the PR rather than author of the commit, as the PR is the responsibility of the author).

Fairly simple algorithm. Check status failures as per normal, check to see if VB posted a comment after that date, if not then post comment informing a failure has occurred. If comment already exists, ignore (to prevent spam).

lurch:
Does this partially conflict with #88 ?

hedss:
No, #88 was about reducing spam where pending items were flagged as failures. If you've used VB in the last week you'll note that it's *extremely* quiet now, unless something does actually go wrong.

This issue is about ensuring that, should a status failure actually occur, then the original author of the PR is explicitly pinged so they know to come back to it (there are quite a few engineers who don't look at a PR again until it's been reviewed, and reviewers who won't look at a PR if it fails, so this reduces deadlock).

As mentioned above, the comments will be date checked. Should a status failure have occurred and the author have been pinged *since* that failure, then it will be ignored.

lurch:
> so this reduces deadlock

:tada:

PR #123, opened 2017-06-05
Modify checks on PR when statuses change; inform author that changes are required if any fail

hedss:
Previously checking of PRs was restricted to Versionist commits, and then general
failures on merge.
We now specifically check all statuses on PR modification, and if finding a failling
required status check:
* Check comments to see if any were made post-last commit, and if any comments were bots.
* If not, then we inform the author of failing status checks.

This commit also ensures the `docs/.nojekyll` file is always added post documentation
creation (to ensure Github pages works).

Also updated our TSlint config.

Connects to: #99

Change-type: patch


lib/bots/versionbot.ts
```diff
@@ -401,7 +401,8 @@ export class VersionBot extends ProcBot {
      * Checks the newly opened PR and its commits.
      * 1. Triggered by an 'opened', 'synchronize' or 'labeled' event.
      * 2. If any PR commit has a 'Change-Type: <type>' commit, we create a status approving the PR.
-     * 3. If no PR commit has a 'Change-Type: <type>' commit, we create a status failing the PR.
+     * 3. If no PR commit has a 'Change-Type: <type>' commit (or any other statuses have failed),
```
lurch: "any other statuses" - is this actually checking only required statuses?

lib/bots/versionbot.ts
```diff
@@ -492,22 +497,55 @@ export class VersionBot extends ProcBot {
...
+                        number: pr.number,
+                    },
+                    method: this.githubApi.issues.getComments
+                }).then((comments: GithubApiTypes.Comment[]) => {
+                    // Check for any comments that come *after* the last commit
+                    // timestamp and is a Bot.
+                    let foundBotComment = false;
+                    for (let comment of comments) {
+                        if ((Date.parse(lastCommitTimestamp) < Date.parse(comment.created_at)) &&
+                            (comment.user.type === 'Bot')) {
```
lurch: In case we end up with multiple ProcBots, all commenting on PRs about different things, would it be possible to tighten this check to only check for comments by VersionBot?

lib/bots/versionbot.ts
```diff
@@ -492,22 +497,55 @@ export class VersionBot extends ProcBot {
...
+                        return Promise.resolve();
+                    }
+
+                    // Now we ping the author, telling them something went wrong.
                     return this.githubCall({
                         data: {
-                            body: `@${event.cookedEvent.data.sender.login}, please ensure that at least one commit ` +
-                                'contains a `Change-Type:` tag.',
+                            body: `@${event.cookedEvent.data.sender.login}, status checks have failed ` +
+                                'for this PR. Please make appropriate changes and recommit.',
```
lurch: Would there be any benefit to this specifying _which_ status checks have failed? Or is the expectation that the user will have to check that themselves by visiting the github PR page?

lib/bots/versionbot.ts
```diff
@@ -492,22 +497,55 @@ export class VersionBot extends ProcBot {
...
+                    if (foundBotComment) {
+                        return Promise.resolve();
+                    }
+
+                    // Now we ping the author, telling them something went wrong.
                     return this.githubCall({
                         data: {
-                            body: `@${event.cookedEvent.data.sender.login}, please ensure that at least one commit ` +
-                                'contains a `Change-Type:` tag.',
+                            body: `@${event.cookedEvent.data.sender.login}, status checks have failed ` +
```
lurch: I guess a `const prAuthor = event.cookedEvent.data.sender.login;` might make this a tiny bit more readable? ;)

lib/bots/versionbot.ts
```diff
@@ -492,22 +497,55 @@ export class VersionBot extends ProcBot {
...
+                    }
+                    if (foundBotComment) {
+                        return Promise.resolve();
+                    }
+
+                    // Now we ping the author, telling them something went wrong.
                     return this.githubCall({
                         data: {
-                            body: `@${event.cookedEvent.data.sender.login}, please ensure that at least one commit ` +
-                                'contains a `Change-Type:` tag.',
```
lurch: Is this "error warning" now visible somewhere else?

lib/bots/versionbot.ts
hedss: I've not done this for a few reasons:
1) If anyone else comments on the PR after a commit, it's going to show up in your notifications anyway, which was the point of doing this in the first place
2) Comments aren't like commits, in that you can't pick a username for the bot to commit as (which is how we check those elsewhere). This means that you *have* to know the names of the Github Apps they're running as. This has massive potential to get screwed up (we have several instances of VB running under different App names).

lib/bots/versionbot.ts
hedss: Yes, it only checks required statuses for the branch.

lib/bots/versionbot.ts
hedss: Nope. The point of this PR (and the original issue) is to alert those who require to know, not what may have failed. It doesn't matter *what* failed, as long as the author/maintainer realises the PR's halted for some reason.

lib/bots/versionbot.ts
hedss: It's now part of the implicit status failure. The Commit and PR guide has the styles to follow for commits, it's pretty obvious what's happened if there's a big red cross by `Versionist` and it's telling you to make changes, IMHO.

lib/bots/versionbot.ts
```diff
@@ -414,6 +414,7 @@ export class VersionBot extends ProcBot {
         const head = event.cookedEvent.data.pull_request.head;
         const owner = head.repo.owner.login;
         const name = head.repo.name;
+        const author = event.cookedEvent.data.sender.login;
```
lurch: Is this the author of the whole PR, or the author of the most recent commit which triggered this update?
(and I wonder if it should ping both, if they're different people?)

hedss: Author of the most recent commit. Maybe it should be both. Hmm.

lib/bots/versionbot.ts
lurch: Okay, so will `versionist` then tell you about a missing `change-type:` tag?

hedss: I think something's got lost here. The `Versionist` status check failing will bring up the above message. Do we still want there to be an explicit message about the `Change-Type` tag? Between the failure and the guide, and the access to FlowDock, I think it's pretty explicit. It also means that if other people want to take this and apply it to a completely different `versionist` config, it still works without misleading. I'd rather that than keep a hardcoded message in.

lurch: "The Versionist status check failing will bring up the above message."

Yeah, that's exactly what I was asking about, sorry for the confusion.

lib/bots/versionbot.ts
hedss: Fixed to be author and committer (if the two are the same, then it only does one ping to the author).

lib/bots/versionbot.ts
```diff
@@ -468,6 +470,7 @@ export class VersionBot extends ProcBot {
             }
             if (commits.length > 0) {
```
lurch: Just out of curiosity: is it actually possible to create a PR with 0 commits?

docs/classes/_bots_versionbot_.versionbot.html
```diff
@@ -380,7 +380,7 @@ <h4 class="tsd-returns-title">Returns <span class="tsd-signature-type">ServiceLi
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
-									<li>Defined in <a href="https://github.com/resin-io-modules/resin-procbots/blob/master/lib/bots/versionbot.ts#L754">bots/versionbot.ts:754</a></li>
+									<li>Defined in <a href="https://github.com/resin-io-modules/resin-procbots/blob/master/lib/bots/versionbot.ts#L803">bots/versionbot.ts:803</a></li>
```
lekkas: nit: should we suppress docs/* diffs? If so we could do it with a `.gitattributes` on the root dir:

```
.gitattributes
-------------

docs/* linguist-generated=true
```

hedss: I hadn't suppressed them up to now as documenting ProcBots is one of the open issue items, and this shows updates. Although, in this case, it's just line numbers changing as there are very few actual documentation comments changing. It's a tricky one.

hedss: It's autogenerated anyway, I've added one.

.gitattributes
```diff
@@ -0,0 +1 @@
+docs/* linguist-generated=true
```
pimterry: Didn't know this existed, very cool.

lib/apis/githubapi-types.d.ts
```diff
@@ -20,20 +20,7 @@ limitations under the License.
...
-                owner: {
-                    login: string;
-                }
-            };
-            sha: string;
-        };
-        html_url: string;
-        url: string;
-    };
+    pull_request: PullRequest;
```
pimterry: Should we do this for PullRequestReviewEvent below too, while we're here?

build/apis/githubapi-types.d.ts
```diff
@@ -78,16 +65,30 @@ export interface StatusEvent {
...
 // Github API -----------------------------------------------------------------
 // These provide the current bare minimum definitions for child Procbots working with them.
+export interface Comment {
+    body: string;
+    user: {
+        login: string;
+        type: string;
+    };
+    created_at: string;
+    updated_at: string;
```
CameronDiver: Why are we not using `Date` objects to store this? As this is going to be a library, seems like that would make the most sense. Gonna leave it up to you though obviously.

lib/bots/versionbot.ts
```diff
@@ -492,22 +503,60 @@ export class VersionBot extends ProcBot {
...
+                // comment telling the author of a failure. If there has, then
+                // it's implicit that we made a comment.
+                const lastCommitTimestamp = lastCommit.commit.committer.date;
+                return this.githubCall({
+                    data: {
+                        owner,
+                        repo: name,
+                        number: pr.number,
+                    },
+                    method: this.githubApi.issues.getComments
```
CameronDiver: Is this suscetible to the pagination issues that @sqweelygig has also been experiencing?

lib/bots/versionbot.ts
```diff
@@ -492,22 +503,60 @@ export class VersionBot extends ProcBot {
...
+                    // timestamp and is a Bot.
+                    let foundBotComment = false;
+                    let warningUsers = '';
+                    for (let comment of comments) {
+                        if ((Date.parse(lastCommitTimestamp) < Date.parse(comment.created_at)) &&
+                            (comment.user.type === 'Bot')) {
+                            foundBotComment = true;
+                            break;
+                        }
+                    }
```
CameronDiver: This could be 
```
const foundBotComment = _.any(comments, (comment) => {
   return (Date.parse(lastCommitTimestamp) < Date.parse(comment.created_at))
        && comment.user.type === 'bot
})
```
Which seems a lot more succinct.

CameronDiver: And if the dates were stored as `Date` objects, we could drop the parses inside.

lib/bots/versionbot.ts
```diff
@@ -492,22 +503,60 @@ export class VersionBot extends ProcBot {
...
+                    if (foundBotComment) {
+                        return Promise.resolve();
+                    }
+
+                    // Now we ping the author, telling them something went wrong.
+                    warningUsers = `@${author}, `;
+                    if (author !== committer) {
+                        warningUsers += `@${committer}, `;
+                    }
```
pimterry: Would be nice to refactor this down a little for clarity, to something more like:

```typescript
const lastCommitTimestamp = lastCommit.commit.committer.date;

getBotComments().then((botComments) => {
  if (Date.parse(_.last(botComments).created_at) > Date.parse(lastCommitTimestamp)) {
    let warningUsers = `@${author}, `;
    if (author !== committer) {
      warningUsers += `@${committer}, `;
    }
    commentOnPr(pr, `${warningUsers}status checks...`), 
  }
});
```

Definitely feels like there's quite a few places where we're using `githubCall` directly, which we could wrap up nicely like this to make everything much clearer and more readable.

build/apis/githubapi-types.d.ts
hedss: These are the types returned from the Github SDK API. Whilst it defines the object types we pass into it, it unfortunately doesn't do the same with return objects. Essentially this entire file is to make the returned objects from that API type safe (I am dreading turning this into a general purpose NPM module, as I'm going to spend a very boring day filling out a lot of these structures...).

lib/bots/versionbot.ts
hedss: A good spot! Unfortunately we have no control over the format of the date (it's a Zulu time string).

lib/bots/versionbot.ts
hedss: That's a good point. I think if it's sub-100 comments, then it should be ok. However, I think I maybe the Github ServiceEmitter should probably handle this automatically. Will do this for this PR.

build/apis/githubapi-types.d.ts
CameronDiver: Ah ok, so if you wanted to go down that route you would have to create an entire layer of richer data structures on top, got it. Ignore me then 👍

lib/bots/versionbot.ts
CameronDiver: Yeah completely agree, just thought it might be worth noting.

lib/apis/githubapi-types.d.ts
hedss: Yep, shall be done!

lib/bots/versionbot.ts
hedss: `githubCall` is going to vanish completely, in fact if you look at https://github.com/resin-io-modules/resin-procbots/pull/112 that's the point of that PR :)

I'll go through and try and make things a bit simpler.

lib/bots/versionbot.ts
lurch: I guess the `lastCommitTimestamp` could be parsed outside of the `_.any`? And probably makes sense to swap around the order of the `&&` arguments (i.e. check `comment.user.type === 'Bot'` before parsing Dates) to allow for more rapid short-circuiting?

CameronDiver: Nice catch @lurch

lib/bots/versionbot.ts
hedss: I've reduced some extra calls that were no longer required. I think I'm going to hold it there, though. I've some ideas of how to make this better, but I want the `dispatchToEmitter` change in first, else it's duplicated work.

lib/bots/versionbot.ts
hedss: Done!

lib/bots/versionbot.ts
hedss: @CameronDiver Done. If you look at `lib/services/github.ts` you'll see the `sendData()` call now automatically deals with paged requests, but is still bw compatible. Could you give it a brief once-over, please (it is tested)?

Issue #130, opened 2017-06-07
NotifyBot should accept both github 'closing' keywords and waffleboard conventions


lekkas:
see: https://www.flowdock.com/app/rulemotion/r-process/threads/xqC1B_QZwkr3fV6fqRCGNm2LdNQ

When a PR references an issue in the same repo, notifybot should accept both:

- Github keywords that close the issue (https://help.github.com/articles/closing-issues-via-commit-messages/)
- Waffleboard `connects to #` convention

The latter will be removed when the *awesome* NotifyBot renders waffleboard and its workflow obsolete

PR #131, opened 2017-06-07
Accept both github 'closing' keywords and waffleboard conventions when linking to issues

lekkas:
Change-Type: patch
Closes: #130 

By the way, I'm totally open on simplifying the regexp to include the exact keywords, even if that means repeating the shared stems of the words cc @hedss


lib/bots/notifybot.ts
```diff
@@ -85,7 +85,9 @@ interface TopicIssues {
 
 const NotifyBotPort = 8399; // Not a listed registered port
 
-const ConnectRE = /connects-to:[\s]+(#[0-9]+)/i;
+// https://help.github.com/articles/closing-issues-via-commit-messages/
+// https://help.waffle.io/faq/waffle-workflow/use-waffles-connect-keyword-to-connect-prs-to-issues
+const IssueReferenceRE = /(?:close[sd]?:?|fix(?:e[sd]{1})?:?|resolve[sd]?:?|connect(?:s|ed)?[\s]+(?:to)?)[\s]+(.*)/i;
```
hedss: I think Github actually requires the '#' to reference this issues, so it needs to be tested for, and in fact NotifyBot differentiates HQ from local issues by the '#', so this RegExp needs to capture the '#' as well.
Length of the RegExp doesn't bother me. :)

hedss:
@lekkas Ping! :)

lib/bots/notifybot.ts
lekkas: @hedss adding '#' since NotifyBot needs it.

lekkas:
@hedss I guess we'd better go with a `rebase and merge` since this is on a already WIP branch right?

hedss:
@lekkas Yep, that works for me! I don't think I've done anything on it since, but should be fine. :+1:

lekkas:
cool, yeah mostly to avoid having a merge commit in your branch ;)

hedss:
:D Thank you!

PR #125, opened 2017-06-05
Fix broken documentation links in README.md

lurch:

PR #133, opened 2017-06-12
Fix architecture documentation links in README.md

lekkas:
change-type: patch


lurch:
Hah #125 ;-)

But when clicked on, these links still get displayed as raw markdown, rather than being properly formatted :-/

hedss:
@lurch Yes, they will because of Jekyll. I spent some time Friday trying to get the GH pages emulation stuff to work locally, and had it working perfectly. Uploading it to GH broke it. So I'm going to have to spend more time to track this down properly. :(

Issue #115, opened 2017-06-01
SyncBot: Documentation


sqweelygig:
Confirm and create as necessary documentation around how SyncBot works, and how to set it up.  Probably including a small slide deck.
I realise this issue is quite meta, but it is here as it is still a task that needs tracking and remembering.

sqweelygig:
This should include migration of the prototyped, in-repo documentation, to process, as described here (https://github.com/resin-io-modules/resin-procbots/pull/111)

PR #138, opened 2017-06-14
Documents an FAQ for SyncBot

sqweelygig:
Closes: #115 

* Improves on the start made with the SyncBot FAQ
* For copying into the process repo

Change-Type: patch


hedss:
Adding @brownjohnf as he had to configure it running, and therefore has some more insight into it.

docresources/SyncBot/syncbot-readme.md
```diff
@@ -2,10 +2,120 @@
...
+
+Flowdock will search the 1-1 history of a user for a phrase that matches "My
+{service} token is blah", returning blah
+
+#### Error reporting
+
+Errors, where possible, are reported in the originating thread as whispers.
+
+## Configuration
+
```
hedss: This needs to be filled in with the appropriate environment variables SyncBot needs to operate, and what their values should be.
Given there's an example script too, it's probably worth mentioning it and where it is in the repository so people can look at it.

docresources/SyncBot/syncbot-readme.md
```diff
@@ -18,7 +128,7 @@ A little Bot, using the ProcBot framework, that links communication services
     * Uses front's `comment`
     * Without syntax, private is assumed
 
-## Your Configuration
+### Your Configuration
 
 * Ensure your Front, Discourse & Flowdock usernames match
```
hedss: Do we now have a hardlink for the docs you wrote for process? If so, I'd suggest adding the link here.

docresources/SyncBot/syncbot-readme.md
brownjohnf: Here's the config schema from the keyframe:

```
syncbot:
  - name: MESSAGE_CONVERTOR_PRIVATE_INDICATORS
    value: '[" "]'
  - name: MESSAGE_CONVERTOR_PUBLIC_INDICATORS
    value: '[" ", "%"]'
  - name: MESSAGE_SERVICE_PORT
    value: '4567'
  - name: SYNCBOT_GENERIC_AUTHOR_ACCOUNTS
    value: '{"flowdock":{"user":"SyncBot","token":""},"discourse":{"user":"SyncBot","token":""}, "front":{"user":"SyncBot","token":""}}'
  - name: SYNCBOT_SYSTEM_MESSAGE_ACCOUNTS
    value: '{"flowdock":{"user":"SyncBot","token":""},"discourse":{"user":"SyncBot","token":""}, "front":{"user":"SyncBot","token":""}}'
  - name: SYNCBOT_HUB_SERVICE
    value: 'flowdock'
  - name: SYNCBOT_DISCOURSE_CONSTRUCTOR_OBJECT
    value: '{"instance":"forums.resin.io","token":"","username":"SyncBot"}'
  - name: SYNCBOT_FLOWDOCK_CONSTRUCTOR_OBJECT
    value: '{"organization":"rulemotion","token":""}'
  - name: SYNCBOT_FRONT_CONSTRUCTOR_OBJECT
    value: '{"inbox_channels":{"":"","":""},"token":""}'
  - name: SYNCBOT_MAPPINGS
    value: '[[{"service":"discourse","flow":""},{"service":"flowdock","flow":""},{"service":"front","flow":""}],[{"service":"flowdock","flow":""},{"service":"front","flow":""}]]'
```

brownjohnf: Also, in general, I think having JSON configuration values is generally bad practice. I'm not sure how to get around it here, but we've actually got 3 tokens, repeated 3 times in different places, with keys that don't really tell you what should be set there. It might be better to split things such that you have something more like:

```
ENABLED_SYNCS=flowdock,front,discourse
DISCOURSE_USER=SyncBot
FRONT_USER=SyncBot
FLOWDOCK_USER=SyncBot
FRONT_CHANNELS=inb_blah:inb_foo,inb_bar:inb_baz
```

etc. That would have saved a lot of time getting this deployed, I think, but I don't know what the internals look like. I guess my general rule of thumb is that configuration should either be a file, or something you could sanely pass on the command line (the origin of env vars in the first place).

hedss: In the situation where we don't need obfuscation, I don't think files is a bad idea, and in this case, there could actually be a `.procbot.yml` config for SyncBot that lives in `lib/bots/syncbot/.procbot.yml` which SyncBot could load for *everything that is publicly admissible*. The problem still comes down to what needs to be opaque (and do channel names, front inbox names, etc. count as this? Certainly it lets potential attackers know that we use particularly named channels on specific services). For these, we can't easily have a config file, as it'd need to be pulled in from elsewhere. Where? :)

sqweelygig:
Oh bugger, forgot to rebase this one.  @hedss Sorry!

brownjohnf:
For public repos, we don't want to commit use-specific configuration. I
generally think that all configuration should be env var + sensible
defaults. A committed .procbots.yml means anyone wanting to use this repo
to deploy their own bot has to fork it. That seems strange.

hedss:
@brownjohnf A good point, not entirely sure what I was thinking.

PR #147, opened 2017-06-16
KeyframeBot is designed to automate the process of validating keyframes and deploying them to an environment

hedss:
* New PRs for product keyframes in the product keyframe repo are validated. They then follow the
  usual route for merging to `master`.
* Any valid merged keyframe can be promoted from `master` to a chosen environment:
  * As long as the promoter is a valid resin.io admin user
  * A version of the keyframe given is valid
  * The given environment is valid
  Keyframes are promoted using the in-built HTTP endpoint, supplying valid credentials and
  a JSON body giving the relevant details. The credentials must be a JWT returned from a
  successful login to resin. An example of a deployment POST is:
  ```
  curl -XPOST http://localhost:7789/deploykeyframe -H 'Authorization: token <token>' \
  -H 'Content-Type: application/json' -d '{"version": "v0.0.1", "environment": "myEnv"}'
  ```

A product keyframe is designated using the KEYFRAMEBOT_PRODUCT_REPO environment variable,
the environments to use as a JSON object in the KEYFRAMEBOT_ENVIRONMENTS environment variable.
Github service variables are as standard.

There is more still to do, but this is a good starting point.

Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #146


hedss:
@brownjohnf  `lib/bots/keyframebot.ts` -> Load diff
It's large so, helpfully, Github has decided to not actually render it by default when you look at the changed files. I hate that GH does this, it means that it's so easy to miss actual, new, substantial code.

There's probably a config to change this somewhere...

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,683 @@
...
+     * @returns             A string naming the branch created on fulfilment.
+     */
+    private createNewEnvironmentBranchCommit = (branchDetails: DeploymentDetails): Promise<string> => {
+        const owner = branchDetails.owner;
+        const repo = branchDetails.repo;
+        const keyframe = branchDetails.keyframe;
+        const environment = branchDetails.environment;
+        const version = branchDetails.version;
+        const user = branchDetails.username;
+        const branchName = `${user}-${version}`;;
```
lurch: Should the `branchName` include the `${environment}` too? Otherwise an error might get thrown if the user tries pushing the same version to two different environments? (or perhaps I'm just misunderstanding this)

EDIT: Ahhh, I think I now see that there's a 1:1 mapping of environments to repos (i.e. each environment is hosted by separate repos). So should KeyframeBot at startup check that two separate environments don't map to the same repo?

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,683 @@
...
+        }).return(branchName);
+    }
+
+    /**
+     * Reports an error to the console and as a Github comment..
+     *
+     * @param error The error to report.
+     */
+    private reportError(error: GithubError | HTTPError): void {
+        // Log to console.
```
lurch: Comment above says "to the console and as a Github comment" but comment here just says "Log to console" ?

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,683 @@
...
+            return this.dispatchToEmitter(this.githubEmitterName, {
+                data: {
+                    owner,
+                    repo,
+                    sha: branchSha,
+                },
+                method: this.githubApi.gitdata.getTree
+            });
+        }).then((tree: GithubApiTypes.Tree) => {
+            // Find the write entry in the tree for the keyframe file.
```
lurch: Is "write entry" correct or was it supposed to be "right entry" ?

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,683 @@
...
+            // Get the right environment.
+            const envRepo = this.environments[environment];
+            if (!envRepo) {
+                throw new HTTPError(404, 'Passed environment does not exist');
+            }
+
+            // Prep for the environment PR.
+            const splitRepo = envRepo.split('/');
+            owner = splitRepo[0];
+            repo = splitRepo[1];
```
lurch: Given how tightly coupled `owner` and `repo` are (without one, the other makes no sense), perhaps there should be a class encapsulating both of these into a single object, to make it easier to pass around? (and I guess this would apply to _all_ the ProcBot stuff, not just KeyframeBot)

Or perhaps that's an over-optimisation/complication? \*shrug\*

(And such an object should also provide automatic stringification too - this file alone has 7 occurrences of `${owner}/${repo}`!)

lib/bots/keyframebot.ts
hedss: It really doesn't matter if they do, it would just act as an alias. :)

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,683 @@
...
+//    curl -XPOST http://<server>:7788/deploykeyframe -H 'Authorization: token <blah> \
+//    -H 'Content-Type: application/json' -d '{"version": "v4.4.0", "environment": "test"}'
+// 5. If the version if valid, and the user is confirmed as a resin.io admin, then a new branch
+//    in the given environment is created based off `master`, and the keyframe is committed to it.
+//    (Branches are named `<user>-<keyframeVersion>`).
+// 6. Finally, a new PR is created in the environment repository, ready for review.
+//
+// TBD:
+//  - Test deployment on opening/update of environment PR to create a status check.
+//  - Extended linting of keyframes along with variables in an environment PR
```
hedss: I think they're separate problems for the moment.

lib/bots/keyframebot.ts
hedss: The reason I've not done this is to allow simplified passing of data to objects. Typescript doesn't require values for keys if the values are the same name as keys:
```typescript
interface MyStructure {
    owner: string;
    repo: string;
}

const intInst = {
    owner: 'theOwner',
    repo: 'theRepo',
};

passInterfaceToMyMethodThatTakesOwnerAndRepoAsPartOfTheStructure({
    ...,
    owner: intInst.owner,
    repo: intInst.repo
    ...,
});
```
However, if we just define owner and repo as local values:
```typescript
const owner = 'theOwner';
const repo = 'theRepo';

passInterfaceToMyMethodThatTakesOwnerAndRepoAsPartOfTheStructure({
    ...,
    owner,
    repo,
    ...,
});
```
I personally feel this is cleaner.

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,683 @@
...
+                throw new HTTPError(404, 'Passed environment does not exist');
+            }
+
+            // Prep for the environment PR.
+            const splitRepo = envRepo.split('/');
+            owner = splitRepo[0];
+            repo = splitRepo[1];
+
+            // Ensure that the version of the keyframe specified actually exists.
+            return this.dispatchToEmitter(this.githubEmitterName, {
```
hedss: No, if the version is invalid then the very act of attempting to retrieve the keyframe itself will result in the following error from Github:
```
No commit found for the ref refs/tags/v0.0.5
```
Which I believe is pretty self-explanatory.

lib/bots/keyframebot.ts
lurch: Fair enough. I guess it depends on the ratio of the number of methods that take `owner` and `repo` as separate parameters (which I guess is all the 'raw' GH API calls) to the number of methods that could be modified to take a single OwnerRepoObject (potentially every other ProcBot method). I guess it also depends on the level of abstraction that you're intending to target with the ProcBots framework.
\*shrug\*

lib/bots/keyframebot.ts
lurch: Ah cool. I obviously had no idea what error would be returned from Github ;-)

hedss: Yeah, it's not obvious. I'll actually add a comment to that effect. :)

lib/bots/keyframebot.ts
hedss: Yep, all the GH API ones take `owner` and `repo` separately, which is the reason it's not in an interface. (Actually, many VB calls for example do pack them into an interface, but then unpack them for exactly this reason.)

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,684 @@
...
+        // New Express app. We'll reuse it in the GH SL.
+        this.expressApp = express();
+        if (!this.expressApp) {
+            throw new Error("Couldn't create an Express application");
+        }
+
+        // Add body parser.
+        this.expressApp.use(bodyParser.urlencoded({ extended: true }));
+        this.expressApp.use(bodyParser.json());
+        this.expressApp.post(DeployKeyframePath, this.deployKeyframe);
```
lurch: I've never used expressApp before, but does this line mean it'll only listen to POST requests, and ignore GET requests?

Is there (or should there be) anything in the ProcBots framework to ensure that endpoints only listen to HTTPS connections and ignore HTTP connections? (or does that fall more under the deployment side?) Or are these bots 'safe' to listen on HTTP as well? Are any of the endpoints vulnerable to replay attacks? (which I guess theoretically might allow somebody to downgrade the version of a keyframe?)

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,684 @@
...
+            // We now go ahead and:
+            // 1. Create a new branch for this, create it from the version passed
+            // 2. Commit the keyframe to that branch
+            // 3. Open a new PR pointing to that branch. Any relevant reviewers can be set (when it works) from
+            //    a `.procbot.yml` config in the env repo.
+
+            // Create a new branch and commit the keyframe to it.
+            deployDetails = {
+                keyframe: keyframeFile,
+                username: decodedToken.username,
```
lurch: I guess this only works if the username decoded from the POSTed token matches a GitHub username? Is that guaranteed to be the case? Should that condition be checked for?

(or is the username here for display-purposes only, and so wouldn't matter if it doesn't match a GH user?)

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,684 @@
...
+    environments: string;
+}
+
+// KeyframeBot listens for PullRequest and PullRequestReview events on both product
+// and environment repositories. The general flow is:
+// 1. User creates a PR to update the product keyframe.
+// 2. KeyframeBot lints the PR to ensure the keyframe is valid. It sets a status
+//    on the PR (pass/fail), along with an error if one exists.
+// 3. Usual PR flow is carried out, and eventually keyframe is merged to `master`.
+// 4. At some future point, a user requests that the keyframe is deployed to an environment.
```
lurch: I guess in theory steps 2 and 4 here could be performed by entirely separate bots, and they're only lumped together here for convenience?

lib/bots/keyframebot.ts
```diff
@@ -0,0 +1,684 @@
...
+                    repo,
+                    sha: branchSha
+                },
+                method: this.githubApi.repos.getCommit
+            });
+        }).then((lastCommit: GithubApiTypes.Commit) => {
+            // We have new tree object, we now want to create a new commit referencing it.
+            return this.dispatchToEmitter(this.githubEmitterName, {
+                data: {
+                    message: `Update keyframe from product version ${version} on behalf of Resin user ${user}.`,
```
lurch: Should this be `Resin admin` rather than `Resin user`, to match line 430?

lib/bots/keyframebot.ts
hedss: This falls under deployment. Almost all of our services listen to HTTP and then get HTTPS routes setup on top of them by a front-facing system (such as nginx).

lib/bots/keyframebot.ts
hedss: The username for the deployment has nothing to do with the Github username of the Resin admin doing the deploy. This is resin only, and the PR opened states as such. The PRs are being opened in environments that, at most, probably only two people have access to, so there should be absolutely no mention of GH names in those repos.

lib/bots/keyframebot.ts
hedss: They could be, but I don't think it would be sensible for them to be.

lib/bots/keyframebot.ts
hedss: Yep!

lib/bots/keyframebot.ts
lurch: With so many variables flying around all over the place ( :wink: ) perhaps renaming the variable from 'username' to 'resinUsername` would make the fact that it's unrelated to GH more obvious.

> there should be absolutely no mention of GH names in those repos.

Ahh, I wasn't sure if maybe the username was being `@`-mentioned in the PR, so that if it _did_ match a GH user, they'd get a notification ping ;)

lib/bots/keyframebot.ts
lurch: Haha, at least some of my comments are useful :wink: :tada:

lib/bots/keyframebot.ts
hedss: Hmm, I'm not convinced by the username renaming, because the token obviously comes from the SDK, and therefore is a Resin user implicitly?
And no, no pinging occurs. And even if it did, they wouldn't be able to see it. :)

lurch: Perhaps in the line `username: decodedToken.username` it's "obvious" that the username comes from Resin and not GH, but what about https://github.com/resin-io-modules/resin-procbots/pull/147/files#diff-f1b921eba16bdedbd08e12a5d6e04595R92 or https://github.com/resin-io-modules/resin-procbots/pull/147/files#diff-f1b921eba16bdedbd08e12a5d6e04595R464 ?  \*shrug\*

> And even if it did, they wouldn't be able to see it. :)

Haha, I did warn you I was a keyframe newbie :laughing:

hedss: I've updated the `DeployDetails` interface accordingly. I don't think it needs anything else.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file
 automatically by Versionist. DO NOT EDIT THIS FILE MANUALLY!
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## v3.2.0 - 2017-07-03
+
+* KeyframeBot is designed to automate the process of validating keyframes and moving them to the chosen environment. [Heds Simons]
```
lurch: Was this changelog entry missing a `KeyframBot: ` prefix?

hedss: It's the first instance of KeyframeBot and is adding it. The prefix is part of the sentence in this case. I could change this to a prefix I guess. I've been using `bot:` because usually it's easier than creating a sentence that has to parse properly with the botname as the first word...

lurch: Don't mind me, I was just being pedantic... ;)

hedss: Wouldn't have it any other way. ;D

PR #151, opened 2017-06-20
Build following recent documentation

sqweelygig:
* Fixes error where sqweelygig forgot to build following documenting

Change-Type: patch


sqweelygig:
okay... wtf

sqweelygig:
Okay, that kinda makes sense.  Master had received a rebuild, which flushed the inconsistency I had introduced out, which made this a no-op, which closed.

Issue #153, opened 2017-06-20
SyncBot: Redact logs


sqweelygig:
Will describe better once fix is in place, details available on request, but for now I want an issue number for the branch.

sqweelygig:
The log output may contain token information, this should be redacted before logging.

PR #155, opened 2017-06-20
Redacts tokens from log output

sqweelygig:
Connects to #153 

* Replaces any token fields present in error reports with 'redacted'
* Acknowledges in typing that the source of an event may have tokens
* Retires now redundant type InterimIds

Change-Type: patch


lib/bots/syncbot.ts
```diff
@@ -295,11 +301,18 @@ export class SyncBot extends ProcBot {
...
         this.logger.log(LogLevel.WARN, error.message);
-        this.logger.log(LogLevel.WARN, JSON.stringify(event));
+        const logEvent = _.cloneDeep(event);
+        if (logEvent.toIds.token) {
+            logEvent.toIds.token = 'redacted';
+        }
+        if (logEvent.sourceIds.token) {
+            logEvent.sourceIds.token = 'redacted';
+        }
+        this.logger.log(LogLevel.WARN, JSON.stringify(logEvent));
```
petrosagg: Can we not make the logger itself redact the sensitive tokens with a `str.replace(new RegExp(_.escapeRegExp(token), 'g'), 'redacted')`?

hedss: I think this makes a huge amount of sense. Cuts down repeated code cruft for this change considerably.

hedss:
As a general comment, I'd like to start putting the name of either the bots affected (or if it's the framework 'ProcBots') as the prefix for each entry that's going to end up in the changelog. So, in this case prefix with `SyncBot: `. However, if we're changing the logging mechanism to carry out the work, then it'd be a general framework change, so would be prefixed `ProcBots: `. I'm just as guilty of not doing this, and we need to start making it more obvious exactly which component has changed. Thanks!

sqweelygig:
Pinging @hedss and @petrosagg about the redaction issue in SyncBot.  I've put a bunch of the utility across into the logger function.

lib/bots/syncbot.ts
```diff
@@ -142,7 +141,10 @@ export class SyncBot extends ProcBot {
     private handleError(error: Error, event: InterimContext): void {
         // Put this on the log service
         this.logger.log(LogLevel.WARN, error.message);
-        this.logger.log(LogLevel.WARN, JSON.stringify(event));
+        const secrets = [];
+        if (event.toIds.token) { secrets.push(event.toIds.token); }
+        if (event.sourceIds.token) { secrets.push(event.sourceIds.token); }
+        this.logger.log(LogLevel.WARN, JSON.stringify(event), secrets);
```
hedss: Maybe:
```typescript
this.logger.log(LogLevel.WARN, JSON.stringify(event),
    _.filter(_.map([event.toIds, event.sourceIds], (thing) => thing.token), (token) => token)));
```
instead of the two ifs would be better, or break the test out into a static method, as it's used by both handlers?
If we're keeping the `ifs`, can we please follow convention in the rest of the framework and do:
```typescript
if () {
    statement;
}```

File: lib/bots/syncbot.ts [hunk 3]
```diff
@@ -295,11 +297,14 @@ export class SyncBot extends ProcBot {
...
-    private logError(error: Error, event: MessengerContext): void {
+    private logError(error: Error, event: InterimContext): void {
         // Do what we can to make this event obvious in the logs
         this.logger.log(LogLevel.WARN, 'v!!!v');
         this.logger.log(LogLevel.WARN, error.message);
-        this.logger.log(LogLevel.WARN, JSON.stringify(event));
+        const secrets = [];
+        if (event.toIds.token) { secrets.push(event.toIds.token); }
+        if (event.sourceIds.token) { secrets.push(event.sourceIds.token); }
+        this.logger.log(LogLevel.WARN, JSON.stringify(event), secrets);
```
hedss: Same here as comment in handleError.

sqweelygig:
reping @hedss

PR #156, opened 2017-06-21
TobyBot: check palindromes in your Github issues!

imrehg:
Adding new demo bot.

### DIFF
--- a/lib/bots/tobybot.ts
+++ b/lib/bots/tobybot.ts
@@ -1,0 +1,194 @@
+/*
+ Copyright 2016-2017 Resin.io
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+// TobyBot listens for Github Issues, and counts the palindromes in them
+//
+// To set up, follow the main guide, and define these respective env vars:
+// TOBYBOT_WEBHOOK_SECRET
+// TOBYBOT_INTEGRATION_ID
+// TOBYBOT_PEM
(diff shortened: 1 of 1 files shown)


lib/bots/tobybot.ts
```diff
@@ -0,0 +1,194 @@
...
+        if (palindromeCount === 1) {
+          countWord = 'palindrome';
+        }
+
+        // Test if the whole issue is a palindrome!
+        let palindromeCombo: boolean = false;
+        const bodySimplified = body.replace(/\W/g, '').toLowerCase();
+        if ( bodySimplified === bodySimplified.split('').reverse().join('') ) {
+          palindromeCombo = true;
+        }
```
lurch: ...which would then simplify this to just `const palindromeCombo = isPalidrome(body.replace(/\W/g, ''));`

lib/bots/tobybot.ts
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+// TobyBot listens for Github Issues, and counts the palindromes in them
+//
+// To set up, follow the main guide, and define these respective env vars:
+// TOBYBOT_WEBHOOK_SECRET
+// TOBYBOT_INTEGRATION_ID
+// TOBYBOT_PEM
+// TOBYBOT_NAME
+// TOBYBOT_EMAIL
```
imrehg: @hedss I guess that's a holdover from the syncbot example and the email setting is not really needed, right? Or I'm double-missing something?

hedss: Sounds like it, yes! I don't see any reason to need it.

hedss:
Currently only holding off reviewing this as VersionBot needs a change to allow it to see forked repos that aren't owned by Resin. Once that work is in, we'll come back to this!

imrehg:
@hedss thanks, noted, and will address the other comments before before continuing too.

imrehg:
@hedss how are bots, would this PR be something that can be revived, or should be dropped? Just as a current status report, I know that there are more important things to do, just to see whether I should keep this around for the future. And not a hurry for any reply. :)

hedss:
@imrehg I don't want to drop it, but it requires the changes to allow non-resin org owned repos. I'm hoping I'll get some time to do this, or someone else will, early next year!

lurch:
Have a safe Christmas, TobyBot :gift: :christmas_tree: :gift: toByboT ,samtsirhC efas a evaH

Issue #27, opened 2017-02-13
VersionBot:  Check to ensure an approved review is by an approved reviewer


hedss:
We've been discussing configuration of ProcBots for a while. We could then, in fact, have a `procbots.yml` file in the root of each repository and include the following section:
```
reviews:
    minimum_approves: 2
    maintainers:
        - hedss
        - lekkas
```
This would then allow, for example, VersionBot to determine that it can only merge when:
* At least one reviewer has approved the PR
* When **only** either 'hedss' or 'lekkas' has added the `procbots/versionbot/ready-to-merge` tag on the PR

lurch:
:+1: to a `procbots.yml` file - it's easy to imagine there'll be lots of things that'll need per-repo (and/or per-branch) configuration.

Given that the labels have been "name-spaced" as e.g. `procbots/versionbot/ready-to-merge`, I wonder if similar "name-spacing" should be used inside `procbots.yml`?

> `minimum_approves: 2` ... At least one reviewer has approved the PR

Huh, typo?

> When **only** either 'hedss' or 'lekkas' has added the `procbots/versionbot/ready-to-merge` tag on the PR

Should at least one of 'hedds' or 'lekkas' have approved the PR too? Or are you making that implicit within the action of them adding the `procbots/versionbot/ready-to-merge` label?  Given that the Github web interface doesn't make it clear who the last person to add a label was (without scrolling through the history of comments) I wonder if it might be better to keep it more explicit as 'at least one maintainer has approved the PR' ?
(but then I'm not that familiar with how labels are used on Github by resin.io, so maybe I'm missing something here? I guess I ought to go and read the associated HQ spec... ;)  )

hedss:
Yes, typo. :)

The assumption is that only one of them needs to have added the label. It doesn't make an assumption about who should review. We usually have many reviewers, but only one or two maintainers of a component.

@lekkas and I have talked at length about a config file, and I'd expect an extra section, such as:
```
reviewers:
  - Page-
  - lurch
  - lekkas
  - hedss
```
etc. for this. The beauty of this is VersionBot could (if we desire) also assign them all on PR open.
The GH API allows VersionBot to determine who has added labels, incidentally.

lurch:
Right, so (one of) the maintainers only needs to add the `procbots/versionbot/ready-to-merge` label, they don't need to have reviewed it too?
[...]

PR #158, opened 2017-06-21
VersionBot/ProcBots: Allow specified reviewer moderation to be enforced; allow services to retrieve configuration files

hedss:
This implements the guarding of PRs via named reviewers and maintainers.
A minimum number of approvals can be specified for a PR in a repository, and various rules will then be invoked to ensure that this target is met before a merge can occur. This is guarded by a `VersionBot` status check that is set informing the user of the number of approved reviews currently acquired and the number required in a `x/y` format.

On opening a new PR, should a configuration file exist for the repository specifying named reviewers/maintainers, `VersionBot` will request that any named user who has not been added to the PR adds themselve.

More details can be found in the sister PR for the Commit and PR guidelines: https://github.com/resin-io/process/pull/87

As part of this work, the use of `.procbots.yml` to store maintainer lists is deprecated. VersionBot now looks for a `repository.yml` file in the root of the repository being operated upon.

Connects-to: #27 
---- Autogenerated Waffleboard Connection: Connects to #27


hedss:
@pimterry @CameronDiver @lekkas Ping! Would you fine folks mind reviewing, please?

lib/bots/versionbot.ts
```diff
@@ -92,20 +91,26 @@ interface ExternalCommand {
     };
 }
 
+enum ReviewState {
+    Approved = 0,
+    ChangesRequired
+};
```
CameronDiver: Do comments need to be considered here?

lib/bots/versionbot.ts
```diff
@@ -143,10 +148,12 @@ const IgnoreLabel = 'procbots/versionbot/no-checks';
  * accordingly.
  */
 export class VersionBot extends ProcBot {
-    /** Github ServiceListener. */
+    /** Github ServiceListener name. */
     private githubListenerName: string;
-    /** Github ServiceEmitter. */
+    /** Github ServiceEmitter name. */
     private githubEmitterName: string;
+    /** Github ServiceEmitter name. */
```
CameronDiver: `name` shouldn't be here right?

lib/bots/versionbot.ts
```diff
@@ -390,6 +409,228 @@ export class VersionBot extends ProcBot {
...
+        const pr = event.cookedEvent.data.pull_request;
+        const head = event.cookedEvent.data.pull_request.head;
+        const owner = head.repo.owner.login;
+        const repo = head.repo.name;
+        let approvedMaintainers: string[];
+        let approvedReviewers: string[];
+
+        // Only when opening a new PR.
+        console.log(event.cookedEvent.data.action);
+        if (event.cookedEvent.data.action !== 'opened') {
```
CameronDiver: If the GH webhooks are running slow, and VB never gets the `opened` event, is there a way it could still perform this task?

lib/bots/versionbot.ts
```diff
@@ -390,6 +409,228 @@ export class VersionBot extends ProcBot {
...
+                },
+                method: this.githubApi.pullRequests.get
+            });
+        }).then((pullRequest: GithubApiTypes.PullRequest) => {
+            // Look at the reviewers, create a list of configured reviewers that were not added
+            // when the PR was opened.
+            const assignedReviewers = _.map(pullRequest.requested_reviewers, (reviewer) => reviewer.login);
+            const configuredReviewers = _.uniq(_.unionWith(approvedReviewers, approvedMaintainers));
+            let missingReviewers = _.filter(configuredReviewers, (reviewer) => {
+                return !(_.find(assignedReviewers, (assignedReviewer) => (assignedReviewer === reviewer)) ||
```
CameronDiver: Super NIT: I think your intent could be a bit clearer here if the `||` was on the following line.

lib/bots/versionbot.ts
```diff
@@ -390,6 +409,228 @@ export class VersionBot extends ProcBot {
...
+                    (reviewer === pr.user.login));
+            });
+
+            // We don't assign the author as a reviewer, if they're in the list.
+            // An App has no ability to create a review request. The best we can do is to create a new
+            // comment pinging those missing in the list to add themselves.
+            if (missingReviewers.length > 0) {
+                let reviewerMessage = '';
+                _.each(missingReviewers, (reviewer) => {
+                    reviewerMessage += `@${reviewer}, `;
```
CameronDiver: I may be completely off-base here, but what would happen if in the `repository.yml` the reviewers are defined with a leading `@`?

lib/bots/versionbot.ts
```diff
@@ -390,6 +409,228 @@ export class VersionBot extends ProcBot {
...
+        });
+    }
+
+    /**
+     * Checks to ensure that the minimum number of approvals for a PR occurs.
+     * Should the conditions be satisfied then a successful status is set, otherwise a failed
+     * state is set.
+     *
+     *
+     *
```
CameronDiver: There's a lot of newlines here :p

lib/bots/versionbot.ts
```diff
@@ -390,6 +409,228 @@ export class VersionBot extends ProcBot {
...
+                    method: this.githubApi.issues.createComment
+                });
+            }
+        });
+    }
+
+    /**
+     * Checks to ensure that the minimum number of approvals for a PR occurs.
+     * Should the conditions be satisfied then a successful status is set, otherwise a failed
+     * state is set.
```
CameronDiver: I didn't see the call in the other changes, will this only trigger in the instance where a user applies the merge label? E.g. versionbot isn't going to add messages stating there isn't enough reviewers on another event?

lib/bots/versionbot.ts
```diff
@@ -390,6 +409,228 @@ export class VersionBot extends ProcBot {
...
+                    number: pr.number,
+                    owner,
+                    repo
+                });
+            }
+            // If the length of the unique list of maintainers and reviewers is less than
+            // the value of `approvalsNeeded`, then this is never going to work.
+            // We only need to test if the `approvedReviewers` list is present
+            if (approvedReviewers &&
+            (_.unionWith(approvedReviewers, approvedMaintainers || [], _.isEqual).length < approvalsNeeded)) {
```
CameronDiver: NIT: I'd move the `_.unionWith` out to another statement, this statement feels quite clunky.

lib/bots/versionbot.ts
```diff
@@ -1278,6 +1504,18 @@ export class VersionBot extends ProcBot {
...
 
     /**
+     * Strip the PR author from a list of user login string.
+     *
+     * @param list          The array of user logins.
+     * @param pullRequest   The pull request to use as a base.
+     * @returns             An array containing stripped user logins, or null should there be no valid users.
+     */
+    private stripPRAuthor(list: string[] | null, pullRequest: GithubApiTypes.PullRequest): string[] | null {
+        const filteredList = (list) ? _.filter(list, (reviewer) => reviewer !== pullRequest.user.login) : null;
```
CameronDiver: Why is `list` parenthesised here?

hedss: Because I have now reached the age where certain code patterns (and style guides) are so hard-welded into my mind I don't even think about them. Shall remove. :)

lib/bots/versionbot.ts
hedss: Waaay too many.

lib/bots/versionbot.ts
hedss: `@` is a specific notification label though, you don't type `@CameronDiver` into the Github login dialogue? I'm reticent to change this, as I don't believe anyone should be adding GH names with this kind of modifier. At worst, it'd do `@@<reviewer>`.

lib/bots/versionbot.ts
hedss: I'm pretty against adding logical operators on the next line, it leaves the previous one hanging with no context as to what's coming. To me, this is far clearer.

lib/bots/versionbot.ts
hedss: If they're running slow, then the event will come later, VB will still add it when it occurs. GH hasn't, as far as I can tell, ever dropped an event, just failed to deliver it in a timely manner. If GH don't deliver events as guaranteed, the entire VB concept unravels.

lib/bots/versionbot.ts
hedss: Yeah, I'll retro comment everything else above, too.

lib/bots/versionbot.ts
hedss: No, this is a new status check, and as such it's registered in the constructor. It doesn't get called from any of the other methods. See the `ghListener.registerEvent` call in the constructor, it's the second registration object.

lib/bots/versionbot.ts
CameronDiver: Sorry that was extremely unclear by me. I mean github comments, for example if somebody is requested for review, and they select `comment` (not `approve` or `request changes`) then github will remove them from the "to-review" list. I'm just wondering if this is something that you need to consider in the reviewer logic?

lib/bots/versionbot.ts
CameronDiver: Ah cool, I'm happy to know that we're not going to get dropped events. Also I didn't even notice that :p

lib/bots/versionbot.ts
hedss: Oh right. No, a comment is considered a 'Well, ok then', just like we're doing with these NITs. :)

CameronDiver: I guess you're right, eventually they'll be required to approve anyway (if they're a maintainer) so it doesn't make a difference. Cool, just wanted to make sure we've considered it :)

lib/bots/versionbot.ts
CameronDiver: I have no problem with that :)

lib/bots/versionbot.ts
CameronDiver: But will it actually ping the person with two `@@` at the front is what I'm worried about. I guess we can test right now :) @@heds

CameronDiver: Ok, looks fine, I'm happy.

lib/bots/versionbot.ts
CameronDiver: Ah, there it is. Thanks man.

lib/framework/procbot.ts
```diff
@@ -83,29 +83,26 @@ export class ProcBot {
...
-    Promise<ProcBotConfiguration | any | void> {
-        // If the path is a string, then we simply try and get from the filesystem,
-        // else we try and get the emitter passed in to read it.
-        let retrievePromise: Promise<any>;
-        if (source === 'fs') {
-            retrievePromise = fsReadFile(<string>location).call('toString');
+    protected retrieveConfiguration(details: ConfigurationLocation): Promise<ProcBotConfiguration | void> {
+        let retrievePromise: Promise<string | void>;
+        if (typeof details.emitter === 'string') {
+            retrievePromise = fsReadFile(details.location).call('toString');
```
lurch: Should this also be checking that `details.emitter === 'FS'` ?

lib/services/github.ts
```diff
@@ -333,6 +334,45 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
         };
     }
 
+    public getConfigurationFile(details: GithubConfigLocation): Promise<string | void> {
```
lurch: Any reason this is passed the whole `details` object, and not just `details.location`?

Issue #157, opened 2017-06-21
VersionBot: Some characters break its git clone


sqweelygig:
@hedss FYI, in [this issue](https://github.com/resin-io/hubot-as-mainbot/pull/285) versionbot failed to check out a branch, probably because it doesn't quote it's branch name in the sh script.

PR #159, opened 2017-06-21
VersionBot: Ensure that repository/branch names are correctly encapsulated to allow handling of all valid git glyphs

hedss:
This ensures that branches such as 'this-is(my)-branch' are correctly dealt with.

Updates TS version and corrects a couple of minor

Connects-to: #157 

Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #157


hedss:
Props to @sqweelygig  for finding this one.

lib/services/github.ts
```diff
@@ -391,7 +392,7 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
                     // If there are some labels, then we process them.
                     const labels = data.response;
                     if (labels) {
-                        const foundLabels: string[] = labels.map((label: any) => {
+                        const foundLabels: string[] = _.map(labels, (label: any) => {
```
CameronDiver: You can just do
```
const foundLabels: string[] = _.map(labels, 'name')

Comment (hedss) [inline lib/services/github.ts]: Yeah, good spot. :)

File: lib/bots/versionbot.ts [hunk 2]
```diff
@@ -769,9 +769,9 @@ export class VersionBot extends ProcBot {
         };
 
         return Promise.mapSeries([
-            `git clone https://${versionData.authToken}:${versionData.authToken}@github.com/` +
-                `${versionData.repoFullName} ${versionData.fullPath}`,
-            `git checkout ${versionData.branchName}`
+            `git clone "https://${versionData.authToken}:${versionData.authToken}@github.com/` +
+                `${versionData.repoFullName}" ${versionData.fullPath}`,
+            `git checkout "${versionData.branchName}"`
```
pimterry: We need to do some much more thorough escaping than this. I think branch names can include nearly anything, definitely including double quotes (which will definitely break this escaping), and we're passing this string directly to the CLI!

The one thing that's holding back branches like `"; rm -rf / #` from being completely viable is that you're not allowed spaces, but I suspect there's still probably a fair few crazy breaking things you can come up with here. `";shutdown` as a branch name will cause some havoc, for starters.

hedss: Actually the right way to do this is to ditch `exec` completely and use `spawn`, I think.

pimterry: Yep, that should work nicely.

lurch: See also https://github.com/ChALkeR/notes/blob/master/Stealing-Travis-secure-variables.md which discusses something vaguely similar.

lurch: @jviotti Would switching from `exec` to `spawn` also fix some of the quoting issues we had on Etcher? Or do our problems run much deeper than that? ;-)

lib/bots/versionbot.ts
```diff
@@ -833,7 +866,10 @@ export class VersionBot extends ProcBot {
             moddedFiles.push(`CHANGELOG.md`);
 
             // Now we get the new version from the CHANGELOG (*not* the package.json, it may not exist).
-            return exec(`cat ${versionData.fullPath}${_.last(moddedFiles)}`).then((contents: string) => {
+            return cliCommand({
+                command: `cat`,
```
lurch: Apologies if this is a dumb question, but why use `cat` to get the file's contents, rather than just opening it directly?

hedss: It's really not a dumb question, and actually I was convinced that I'd changed this in another PR. Having spent time just now looking, I can't find the change. I'll make it here.

lib/bots/versionbot.ts
```diff
@@ -764,14 +769,32 @@ export class VersionBot extends ProcBot {
...
+                command: 'git',
+                args: ['clone', `https://${versionData.authToken}:${versionData.authToken}@github.com/` +
+                `${versionData.repoFullName}`, `${versionData.fullPath}`],
+                workingDir: `${versionData.fullPath}`
+            },
+            {
+                command: 'git',
+                args: ['checkout', `${versionData.branchName}`],
+                workingDir: `${versionData.fullPath}`
+            }
```
pimterry: Could make this a little more readable with a command that builds these objects:

```typescript
buildCommand('git', ['clone', 'https://...'])
```

lib/bots/versionbot.ts
```diff
@@ -764,14 +769,32 @@ export class VersionBot extends ProcBot {
...
+        const cliCommand = (params: ExternalCommand) => {
+            return new Promise((resolve, reject) => {
+                const child = spawn(params.command, params.args,
+                    (params.workingDir) ? { cwd: params.workingDir } : undefined);
+                let stdout = '';
+
+                child.stdout.on('data', (data) => {
+                    stdout += data.toString();
+                });
+                child.addListener('close', () => resolve(stdout));
```
pimterry: Might also be useful to have `stderr` as well?

lib/bots/versionbot.ts
```diff
@@ -764,14 +769,32 @@ export class VersionBot extends ProcBot {
         //     `package.json` file. This means components that don't have one need a custom
         //     `versionist.conf.js` in their root dir. And we need to test to run against it.
         //     It's possible to get round this using a custom `versionist.conf.js`, which we now support.
-        const cliCommand = (command: string) => {
-            return exec(command, { cwd: versionData.fullPath });
+        const cliCommand = (params: ExternalCommand) => {
```
pimterry: Could we put a verb in here? `runCommand`?

lib/bots/versionbot.ts
```diff
@@ -833,7 +866,8 @@ export class VersionBot extends ProcBot {
             moddedFiles.push(`CHANGELOG.md`);
 
             // Now we get the new version from the CHANGELOG (*not* the package.json, it may not exist).
-            return exec(`cat ${versionData.fullPath}${_.last(moddedFiles)}`).then((contents: string) => {
+            return fsReadFile(`${versionData.fullPath}${_.last(moddedFiles)}`).call('toString')
```
pimterry: If you pass an encoding to `readFile` I think it'll come back with a string for you, so you can drop the `toString`

Issue #160, opened 2017-06-21
SyncBot: Enable non-matching usernames


sqweelygig:
These could probably be done in a very similar way to the room lookup, where it tries to find it from the configuration.  Other things to consider could be in-app configuration, if I can think of a way.

sqweelygig:
This has become #257

PR #162, opened 2017-06-22
SyncBot: Allow configuration of usernames that don't match

sqweelygig:
connects to #160

* Provide a function to search a configuration object
* Use this as an option when translating usernames

Chnage-Type: minor


docresources/SyncBot/syncbot-readme.md
```diff
@@ -6,7 +6,11 @@ A little Bot, using the ProcBot framework, that links communication services
 
 ### How do I set up my accounts?
 
-* Ensure your Front, Discourse & Flowdock usernames match
+* Try:
```
hedss: This is not useful for anyone apart from engineers who understand the 'try/catch' paradigm. I would not expect any of our staff who haven't written any code in a relevant language to understand why it's like this. Needs to be altered to something in an obvious English sentence.

Issue #166, opened 2017-06-27
VersionBot: 'invalid username or password' intermittent error


pimterry:
Example: https://github.com/resin-io/resin-api/pull/308#issuecomment-311463051

Worked fine after adding and removing the label a little later (after an unrelated build hiccup), with no other changes. Nothing visible in the github status stats.

hedss:
Seen this occasionally, since we started. I think it's a GHism, which for some reason fails the clone and then happily accepts it after. I'll add a retry into the mechanism for cloning, similar to how the GH Service itself works.

PR #167, opened 2017-06-28
VersionBot: Retry git clone operation should it initially fail due to authentication issues

hedss:
Occasionally there's an issue where the initial attempt to clone doesn't work, where Github
claims authentication is incorrect. It appears on subsequent attempts, this works. To combat
this, VersionBot will now attempt a clone up to three times.

* Should the clone fail, a sanitised comment is made on the PR
* Additionally, the code to use local processes has been moved out of VersionBot and into
  a utility component called 'environment'. The allows a Client Bot to build a command and
  then execute it (and hence the minor semver)
* The gulp file has been cleaned up to remove the need to copy declarations (which we shouldn't
  need)
* The CHANGELOG.md has been retroactively updated, to specify which Client Bot the change
  pertains to (or the overall ProcBots framework)

Connects-to: #166

Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #166


lib/utils/environment.ts
```diff
@@ -0,0 +1,112 @@
...
+ * The command will be attempted based on the value of the `options.retry` property in the
+ * passed Command object. Should no retries be specified, command execution will only
+ * be attempted once.
+ *
+ * @param   params  A Command object detailing the command to execute.
+ * @returns         A Promise containing a string from `stdout`.
+ * @throws          An error containing a message from the failed command's `stderr` output.
+ */
+export function ExecuteCommand(command: Command): Promise<{}> {
+    const retries = (command.options || {}).retries || 1;
```
pimterry: The variable should be called `tries` and be equal to `.retries + 1`, not just `.retries` (or some other equivalent refactoring below).

Currently, this is the total number of times the command is run (which is why `1` is a sensible default). If I pass `{ retries: 0 }` though, I expect the command to be run exactly once, and for `{ retries: 1 }` I expect the command to be run up to two times.

lib/utils/environment.ts
```diff
@@ -0,0 +1,112 @@
...
+        }).catch((err: Error | TypedError) => {
+            // Keep trying until we exhaust retries.
+            retryCount--;
+            if (retryCount > 0) {
+                return Promise.delay(1000).then(callCommand);
+            }
+
+            if (err instanceof Error) {
+                throw new TypedError(err.message);
+            }
```
pimterry: Why do we want to wrap this error in a plain `TypedError`? I thought that was purely used as an abstract class, not something you'd instantiate.

lib/utils/environment.ts
```diff
@@ -0,0 +1,112 @@
...
+                }
+                stdout += data.toString();
+            });
+            child.stderr.on('data', (data) => {
+                if (!stderr) {
+                    stderr = '';
+                }
+                stderr += data.toString();
+            });
+            // If there's any stderr output, we send that instead of stdout.
```
lurch: This should probably be mentioned in the `@returns` line?
Also, what if somebody wants both stderr and stdout as separate strings?
Also also, is there a way to retrieve the command's exit-code? Or does any non-zero error simply throw?

lurch:
Does this only affect clone operations, or are there other Github operations where retrying might be beneficial?

lib/utils/environment.ts
hedss: Anything non-zero throws, at the moment they're living with either the `stderr` or `stdout`. I'll improve this when we get round to proper error handling, too.

hedss:
@lurch Currently this is the only reasonable one to apply this to in VB as it relies on an external source. If any of the others fail, retrying will not help (running `versionist` etc.).

lib/utils/environment.ts
```diff
@@ -16,12 +16,14 @@ limitations under the License.
...
 import { spawn } from 'child_process';
-import TypedError = require('typed-error');
 
 export interface CommandOptions {
     /** The CWD to run the command in. */
     cwd?: string;
-    /** How many retry attempts should occur if the command fails. */
+    /**
+     * How many retry attempts should occur if the command fails. Any value less than
+     * zero will occur in no retries.
```
lurch: 'zero' -> 'one' ?
'will occur' -> 'will result' ?

PR #169, opened 2017-06-29
Move from spaces to tabs in all source files

hedss:
There's a general move to use prettier in Resin source. However, a quick
exploration with it has seen inconsistent indenting and flow. Whilst the
longterm goal is still to use this tool, it needs further investigation
first, so this commit has been carried out using sed (mostly).

Change-type: patch


sqweelygig:
I am going to trust the method, the builder, the Hedley and the skim-read I've just done; because I really don't want to do a line-by-line of this one.

pimterry:
@sqweelygig As an aside (although I too did do a very quick skim to double-check), this is the _perfect_ case for `?w=1`: https://github.com/resin-io-modules/resin-procbots/pull/169/files?w=1

sqweelygig:
Thank you @pimterry, once again your teaching turns your knowledge into a blessing for us all.  (and reveals how many of the non-whitespace fixes are in my code)

Issue #144, opened 2017-06-16
VersionBot: Have the ability to force/require certain commit footer tags


agherzan:
Would be nice to have a configurable option in the local repository config file which will be taken into consideration by versionbot when checking required tags for commits. As well there should be a way of setting if the tag is required globally in the PR (at least once) or in every commit. Useful for example when forcing `changelog-entry`.

This would be a more general solution to https://github.com/resin-io-modules/resin-procbots/issues/142 .

hedss:
We could achieve this with an entry in the `.procbots.yml` files:
```
versionbot:
  required-tags:
    - "signed-off-by"
    - "changelog-entry"
```

agherzan:
@hedss please take into consideration a configuration which would specify if a certain tag is needed in all commits or at least one

Issue #139, opened 2017-06-15
Idea: require reviews from selected people/teams if specific files/folders are touched by PR


lekkas:
This came up after a discussion with @alexandrosm as a way to enforce stricter reviews for anything affecting a component's build/deployment process. For instance, versionbot could explicitly require review from @resin-io/devops if any of

```
- .travis.yml
- circle.yml
- build/
```

has been changed by a PR. This would mitigate the risk of rogue scripts that extract sensitive environment variables (e.g. npm publishing keys)

hedss:
Augment the current `73-27-review-configs` branch to include the following:

```
review_files:
  - files: <array | pattern>
    reviewers: [array of users]
```
Any specific reviewers either adds to (or creates) the `reviewers` list, so they are included on review for any PR that includes changes to those files.

lurch:
But would this prevent the PR from automatically being built and run on the CI services in the first place?

hedss:
@lurch @lekkas So I'd assumed this would just stop it from being merged to `master`, but actually after re-reading the original comment, it does **sound** as if we'd actually be trying to block and PR CI build occuring. Is that right, Kostas?

If so, I'd need to think about this, as I'm not sure we could do so easily. What we'd have to do is something like:
* Keep `master` branch protection as it is now, only merge if other status checks pass
* Ensure VersionBot is the sole status condition as branch protection for the *specific* branch the PR is on
* When reviewers have approved so that the status is successful, VB would then add the other CI systems as required *for the same branch*, which would then carry out builds and report back
* Rest of the process continues as required

I have a bit of a worry, which is by default VB assumes things fail unless otherwise proven (paranoia is goooood!). With this system, VB would be the only thing stopping builds occurring until approvals occur, when the other CI components would be added. On the one hand, no different to how it works now as `master` protection would still be active, on the other hand, VB has to optimistically work forwards when reviews occur.

lurch:
Hmmm, it's something I've never set up myself (so I could be wrong) but I thought that when e.g. adding Travis or Appveyor to a repo, those CI services then automatically build _any_ PRs issued against that repo; and the branch-protection stuff only stops PRs being merged when a status-check fails, it doesn't determine which CI services actually get run against a PR?

Although maybe there's something in the Travis or Appveyor `.yml` files that can be used to tweak this always-build-every-PR behaviour?

[...]

PR #170, opened 2017-06-30
VersionBot: Add ability to specify required tags for a PR in a repository configuration file

hedss:
This functionality allows repository maintainers to determine which footer tags in
a PR are mandatory or optional.
Tags keys are considered case insensitive, and optionally may define how many
occurences should be present in all of the commits for a PR, as well as defining
valid values.

The `change-type` tag is *always* assumed to be present, regardless of other
tag definitions. Should the configuration file redefine the `change-type` tag,
then that definition will override the inbuilt one.

The following shows how to format required tag definitions in the `repository.yml`
configuration:

```
required_tags:
    - name: <tagname>
      occurrence: <number> | 'all'
      values: <regexp>
```

Both the `occurrence` and `values` properties are optional, if present:
* occurence - Specifies the minimum number of commits in the PR that must contain
              the tag. 'all' means every commit in the PR. Should the number of
	      commits in a PR be less than this number, it is assumed that they
	      are valid (as long as those commits contain the tag).
	      Defaults to 1 should it not be defined
* values - A regular expression defining a tags valid values. This regexp should
           be defined as if the entire line containing the tag were present, but
	   should only look for valid values.
	   Defaults to /.*$/ should it not be defined.

As an example, the following is a valid set of definitions for a repository,
and includes the `change-type` tag explicitly.

```
required-tags:
    - name: signed-off-by
      occurrence: all
    - name: changelog-entry
      occurrence: 2
    - name: change-type
      occurrence: 1
      values: /\s*(patch|minor|major)\s*$/i
```

Connects to #144

Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #144


hedss:
Caveat: It's a Friday afternoon and I've got a splitting headache, I may have missed something (though I've tested it pretty thoroughly).

agherzan:
Looks good to me :+1:

lib/bots/versionbot.ts
```diff
@@ -680,69 +701,62 @@ export class VersionBot extends ProcBot {
...
+
+		// Get all the commits on the repo.
+
+		// For each tag we find, we adhere to some rules:
+		//  * all - Every commit in the PR must contain the tag
+		//  * once - At least one commit in the PR must contain the tag
+		//  * <integer> - There must be at least <integer> commits with the tag. If there
+		//                are less commits than the given integer, then all commits must
+		//                include the tag.
+		// Get the reviews for the PR.
```
pimterry: These various comments here seem a bit out of place. Should each of them be next to the sections below that are actually doing what they're describing?

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+
+		// Filter the list of commits to those not made by VB.
+		const commits = _.filter(allCommits, (commit) => commit.commit.committer.name !== process.env.VERSIONBOT_NAME);
+
+		// Get and validate tag configs.
+		for (let tag of tagDefinitions) {
+			// Lower the tag name.
+			tag.name = tag.name.toLowerCase();
+
+			if (tagCounts[tag.name]) {
```
pimterry: If the tag names have to be unique, why not have the configuration be a `{ tagName: { config } }` object, rather than a list where `name` is special? I.e. the config could look like:

```yaml
required_tags:
  connects-to:
    occurance: 1
  change-type:
    occurrence: 1
    values: /\s*(patch|minor|major)\s*$/i  
```

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+				throw new Error(`More than one occurrence of a required footer tag (${tag.name}) found ` +
+					'in configuration');
+			}
+			if (tag.occurrence) {
+				if ((typeof tag.occurrence !== 'number') && (tag.occurrence !== 'all')) {
+					throw new Error(`Invalid occurrence value found for ${tag.name} definition`);
+				}
+			}
+			// Always use a lowered version of the tag.
+			tagCounts[tag.name.toLowerCase()] = 0;
```
pimterry: You've already lowercase'd this above.

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+		let tagCounts: { [key: string]: number } = {};
+		let changeTypeFound = false;
+
+		// Filter the list of commits to those not made by VB.
+		const commits = _.filter(allCommits, (commit) => commit.commit.committer.name !== process.env.VERSIONBOT_NAME);
+
+		// Get and validate tag configs.
+		for (let tag of tagDefinitions) {
+			// Lower the tag name.
+			tag.name = tag.name.toLowerCase();
```
pimterry: If you move to an object, and use `_.mapKeys` up front to lowercase all the names, you can simplify all this. You get to drop this check, assume it's lowercase already everywhere, and drop `changeTypeFound` too. That check later on just becomes:

```typescript
tagDefinitions['change-type'] = tagDefinitions['change-type'] || { ...defaults... };
```

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+					let match = keyRE.exec(footer);
+					while (match) {
+						valueMatches.push(match[1]);
+						match = keyRE.exec(footer);
+					}
+
+					// Match the value.
+					for (let valueMatch of valueMatches) {
+						// Try and match the value returned, if any.
+						const valueFound =  valueMatch.match(valueRE);
```
pimterry: Double space

docresources/VersionBot/versionbot.md
```diff
@@ -115,7 +113,94 @@ procbot:
...
+Both the `occurrence` and `values` properties are optional, if present:
+* `occurence`[optiona] - Specifies the minimum number of commits in the PR that must contain
+                the tag. 'all' means every commit in the PR. Should the number of
+                commits in a PR be less than this number, it is assumed that they
+                are valid (as long as those commits contain the tag).
+                Defaults to 1 should it not be defined
+* `values`[optional] - A regular expression defining a tags valid values. This regexp should
+            be defined as if the entire line containing the tag were present, but
+            should only look for valid values.
+            Defaults to /.*$/ should it not be defined.
```
pimterry: Technically, it defaults to `/.*/` (no `$`). Just being picky though, it doesn't make any difference.

agherzan:
Can `change-type` be overwritten with an entry which would deactivate it completely? Maybe using 0 occurrences?

docresources/VersionBot/versionbot.md
```diff
@@ -115,7 +113,94 @@ procbot:
...
+
+```
+required_tags:
+    - name: <tagname>
+        occurrence: <number> | 'all'
+        values: <regexp>
+```
+
+Both the `occurrence` and `values` properties are optional, if present:
+* `occurence`[optiona] - Specifies the minimum number of commits in the PR that must contain
```
lurch: `[optiona]` -> `[optional]`

docresources/VersionBot/versionbot.md
```diff
@@ -115,7 +113,94 @@ procbot:
...
+                the tag. 'all' means every commit in the PR. Should the number of
+                commits in a PR be less than this number, it is assumed that they
+                are valid (as long as those commits contain the tag).
+                Defaults to 1 should it not be defined
+* `values`[optional] - A regular expression defining a tags valid values. This regexp should
```
lurch: `tags` -> `tag's`

lib/bots/versionbot.ts
```diff
@@ -680,69 +701,62 @@ export class VersionBot extends ProcBot {
...
-			data: {
+		this.logger.log(LogLevel.INFO, `Checking footer tags for ${owner}/${name}#${pr.number}`);
+
+		// Get the configuration, if it exists, for this repo.
+
+		// Get all the commits on the repo.
+
+		// For each tag we find, we adhere to some rules:
+		//  * all - Every commit in the PR must contain the tag
+		//  * once - At least one commit in the PR must contain the tag
```
lurch: Looks like this `once` isn't used anywhere?

lurch:
Just from reading the PR description, the `occurrence` field doesn't seem to make sense :-/  Why specify the minimum number of commits that have to mention the tag, when you obviously have no idea in advance how many commits each PR will contain? IMHO a number doesn't make sense; surely all it needs is settings for "must be on at least one commit" and "must be on every commit" ?  And perhaps there could also be a setting to indicate "must be on one (and at most one) commit"?

Although thinking a bit further, given the regex-checking of tags that this PR introduces, perhaps it'd also be useful to have optional tags too? i.e. An optional tag doesn't have to be on _any_ commits, but if it **is** present, it still has to match that tag's regex?
Incorporating @agherzan 's request too, perhaps `occurrence:` could be set to the strings `none` (this footer-tag must never be present), `optional` (this footer-tag may or may not be present, but if it is the regex has to match), `single` (this footer-tag can appear on at most one commit per PR), `multi` (this footer-tag can appear on one or more commits per PR) and `all` (this footer-tag must appear on every commit in the PR). I've obviously got no idea of the commit-conventions used by the different repos, but if the `occurrence: single` was being used, would there ever be a case for specifying that it has to appear on the `first`-only or `last`-only commit in a PR?

Also, it seems like the regex would be more useful if you could use both `^` and `$`, instead of having to start the regex with just `\s*`? i.e. the current example of `change-type` above would seem to allow a value of `change-type: this is a patch` whereas it might be nice to be able to tighten up the regex to just `/^(patch|minor|major)$/i` ?

Is there a regex that the footer-tag-names themselves should / must adhere to?

And finally one of the examples in the PR description uses `required_tags:` while the other uses `required-tags:`

docresources/VersionBot/versionbot.md
```diff
@@ -115,7 +113,94 @@ procbot:
...
+
+#### Maintainer and Reviewer Configuration
+
+A list of valid maintainers and approved reviewers may be defined to allow VersionBot to vet those Github users who are providing review approval.
+
+The definition for these users in the configuration file is as follows:
+
+````
+minimum_approvals: <number>
+reviewers:
```
lurch: Huh, why has the documentation about this snuck into this PR?

docresources/VersionBot/versionbot.md
lurch: Do you need to escape the `*` with `\*` to stop GH displaying it with weird formatting?

lib/bots/versionbot.ts
```diff
@@ -133,6 +133,24 @@ const RepositoryFilePath = 'repository.yml';
 /** Message sent to required reviewers should they not have been added on a review. */
 const ReviewerAddMessage = 'Please add yourselves as reviewers for this PR.';
 
+/** Defines an interface for configuration defined required tags. */
```
lurch: Might be a tiny bit more readable as "configuration-defined required tags" ?

lib/bots/versionbot.ts
```diff
@@ -133,6 +133,24 @@ const RepositoryFilePath = 'repository.yml';
...
+interface FooterTag {
+	/** Name of the tag, case insensitive. */
+	name: string;
+	/** How many occurrences of the tag must be seen in an PR. */
```
lurch: `an PR` -> `a PR` ;)

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+
+		// For each of the tags, now determine if the threshold occurrence has occurred.
+		let tagResults: MissingTag[] = [];
+		for (let tag of tagDefinitions) {
+			// Lookup the tag in the tag map
+			const tagCount = tagCounts[tag.name];
+			let tagsRequired = 0;
+
+			// Convert all/never to numbers, appropriately. As we pre-validated
+			// tags, we don't need to check for 'never' as that's the only
```
lurch: `never` isn't mentioned anywhere else?

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+				// and have to assume they'll pick a valid tag out.
+				_.each(tagDefinitions, (tag) => {
+					// RE for the key. We want to capture the entire line
+					const keyRE = new RegExp(`^${tag.name}:(.*)$`, 'gmi');
+
+					// We need to compile the RE for the value, if it doesn't already exist.
+					// Strip the RE into a body and flags.
+					let valueRE: RegExp = /.*/;
+					if (tag.values) {
+						// Get flag values, we ignore everything apart from 'i'
```
lurch: Oh?

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+						const regExp = _.split(tag.values, '/');
```
lurch: a) It's a bit confusing naming this `regExp` when it actually stores an array of strings rather than a `RegExp` ;-)
b) I guess this would actually fail if the `tag.values` regex itself contained a `/` character anywhere?

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+				// We should have a line index to join from, now.
+				lines.splice(0, lastLine);
+				const footer = lines.join('\n');
+
+				// For each tag, interrogate the footer and determine if the tag is
+				// present.
+				// We check for a valid instance of a tag. If there are duplicates,
+				// potentially with invalid values, we don't flag it. This is primarily
+				// as we don't know how scripts requiring these tags deal with them
+				// and have to assume they'll pick a valid tag out.
```
lurch: Do we really want to allow duplicate footer-tags within a single commit-message? From what I can tell by reading through the code, this means that a footer-tag with `occurrence: all` would actually be considered valid if a PR had 3 commits with one of the commits having the footer-tag three times and the other two commits not mentioning it at all?

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+					if (tag.values) {
+						// Get flag values, we ignore everything apart from 'i'
+						const regExp = _.split(tag.values, '/');
+						const body = regExp[1];
+						const flags = regExp[2] || '';
+						if (!body) {
+							throw new Error(`Invalid RegExp found for ${tag.name} values`);
+						}
+
+						valueRE = new RegExp(body, flags);
```
lurch: I'm not a TS programmer, but what happens if `body` contains a string that isn't a valid regex-definition?

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+					}
+
+					// Try and match the key.
+					// We want all matches.
+					let valueMatches = [];
+					let match = keyRE.exec(footer);
+					while (match) {
+						valueMatches.push(match[1]);
+						match = keyRE.exec(footer);
+					}
```
lurch: I haven't _quite_ grokked the code fully, but it looks like if a footer-tag is present, but it's value doesn't match the specified regex, then versionbot will report the entire footer-tag as missing? Perhaps it'd be more user-friendly to have a separate error-message for footer-tag-found-but-regex-doesn't-match ?

docresources/VersionBot/versionbot.md
hedss: Because it was missed previously. I'd rather add it here than not have it in at all.

lib/bots/versionbot.ts
hedss: Good catch, it shouldn't be in here.

docresources/VersionBot/versionbot.md
lurch: Ah okay. I wasn't sure if it was deliberate or accidental. (I know it's easy to mix things up when working on multiple PRs at once!)

lib/bots/versionbot.ts
```diff
@@ -352,8 +372,10 @@ export class VersionBot extends ProcBot {
 			});
 
 			// For every one of these, call checkVersioning.
```
hedss: Yup!

lib/bots/versionbot.ts
```diff
@@ -352,8 +372,10 @@ export class VersionBot extends ProcBot {
...
-			return Promise.map(prEvents, (prEvent) => {
-				return this.checkVersioning(registration, prEvent);
+			return Promise.delay(2000).then(() => {
```
hedss: I've noticed a delay in GH being sent a query and then responding. They send us an event telling us the update has occurred, but then a request to get that update (in this case, a comment on a status failure) doesn't actually get the data back. This improves the chance (and in fact, from testing cures in 'normal' circumstances) that we won't get into a situation where VB comments that a status has failed more than once.

lib/bots/versionbot.ts
hedss: Yes, that makes perfect sense. I did say I'd got a headache. ;-)

lib/bots/versionbot.ts
hedss: That's cause it doesn't exist, and this should be taken out.

lib/bots/versionbot.ts
hedss: If it's not in the right format, it's wrong. There's also another pressing tradeoff here, the error message is for a failed status check, and there's a limit to the amount of text you can put in this (this is the reason I've kept the reason for failure in, but don't currently use it). We already have a generic failed status error, I don't want to start adding huge numbers of error comments (at this time), as there are enough people on the fence about whether they want these messages or not. We've hooks now that will let us put this in at a later date if we need to. It's up to the committer to determine why the status check failed, and as we have a Commit and PR guidelines document, it should be pretty quickly determined.

docresources/VersionBot/versionbot.md
hedss: Man, the number of things I'm juggling in OmniFocus and my head, right now! :)

docresources/VersionBot/versionbot.md
hedss: My markdown previewer seemed to cope, hopefully that means GH should?

lib/bots/versionbot.ts
hedss: That's a very good spot. I don't think we should ban duplicates. I've worked at places where the following is a valid commit message:
```
Added: <thing>
Added: <anotherThing>
Replaced: <stuff>
Replaced: <moreStuff>
Removed: <otherStuff>
```
But maybe we should say that all tags must be valid not to fail the check? Actually, yes, let's say that.

lib/bots/versionbot.ts
hedss: The code needs to implicitly do multi and global on the footer anyway, so it's going to ignore any `/mg` variations it finds to do the job properly.

lib/bots/versionbot.ts
hedss: It errors, we get back to the promise chain for the status check and the maintainer of the repo has to ensure that the repo config file is fixed, as VB can't allow anything through that's ambiguous!

lib/bots/versionbot.ts
hedss: Also true, I think I'm going to go with your suggestion of not explicitly stating the RegExp boundaries and assuming everything inside is, with an optional flag for it.

lib/bots/versionbot.ts
lurch: I'd assume that any multiple-occurrence-per-commit tags wouldn't be added to the definitions in `repository.yml` ?
(or if they are, perhaps we should explicitly express those that are allowed to appear multiple times per commit?)

lib/bots/versionbot.ts
lurch: > It errors, we get back to the promise chain

Ah cool, that's the bit I wasn't sure about as I couldn't see any explicit error handling ;)

So does that mean that if somebody commits a PR with an invalid regex in `repository.yml`, it'll only be the **next** PR that gets submitted that will pick up this misconfiguration?

lib/bots/versionbot.ts
lurch: Fair enough, as long as this doesn't end up with committers being confused about why VB is rejecting their PR ;-)

Totally pie-in-the-sky idea, but I wonder if there could be an offline tool that a committer could run locally against their branch, which would use the local repo's `repository.yml` to give them more verbose error reports?

lurch:
Something else I've just thought of - I recall in a `versionist` issue you mentioned that footer-tags had to be at the end of a commit-message, otherwise versionist wouldn't be able to identify them. Does this / should this code also do similar checks / enforcement? (I think the regexes this PR currently uses allow footer-tags to appear anywhere within a commit message?)

hedss:
@lurch No, the code is adapted from the old code to check for `Change-type:` and it does infact look at the entire commit and work backwards to find line breaks to find the footer. So, that's already enforced.

lib/bots/versionbot.ts
```diff
@@ -1550,6 +1571,157 @@ export class VersionBot extends ProcBot {
...
+			// This code will actually let you get away with:
+			//  <header>
+			//
+			//  <footer>
+			// As sometimes a development patch may be self-explanatory in the header alone.
+			const lines = commitMessage.split('\n');
+			const lastLine = _.findLastIndex(lines, (line) => line.match(/^\s*$/) );
+
+			// If there's no match, then at the very least there's no footer, and the commit
+			// is in the wrong format (as there's no text to use in the logs).
```
lurch: > the commit is in the wrong format

I guess it's okay for single commits to be in the 'wrong format', as long as at least one commit in the PR is in the 'correct format'?

hedss: Yep

lib/bots/versionbot.ts
hedss: @flesler has been talking about pre-commit hooks and I'm not sure where that went. I was quite in favour of doing this. :)

lib/bots/versionbot.ts
hedss: So you've come to the point of https://github.com/resin-io-modules/resin-procbots/issues/27 whereby the point of having a maintainer be enforced on a review is to ensure that issues like this don't make it through review. :)

lib/bots/versionbot.ts
flesler: @hedss it's per-project, you can implement it here and not owe the world any explanation :)
You can check [how I did it](https://github.com/resin-io/resin-vpn/issues/25) on the vpn. Note that it has 2 PRs because it started as `precommit` and switched to `prepush` after some complaints. You can keep it on `precommit` if you want, specially if it's only for linting and doesn't even run tests.

lib/bots/versionbot.ts
lurch: ...assuming that the reviewer is able to spot such subtle mistakes ;)

lib/bots/versionbot.ts
lurch: I guess the mindset to adopt when thinking about this PR, is that all the commit-messages in a PR (effectively) get "merged together", and then the `required_tags` checking gets done against this "merged commit message". Which kinda highlights how my idea about only enforcing one footer-tag per PR doesn't really fit within the scope of this PR :)

lib/bots/versionbot.ts
hedss: No I'm not, I'm getting rid of values and adding `never`. :)

lib/bots/versionbot.ts
hedss: @flesler Thank you, thank you!

lib/bots/versionbot.ts
hedss: It's their call. If they feel happy approving code in a potentially volatile file, ultimately it's their responsibility if it fails! :)

lurch: "If it breaks, you get to keep both pieces"

lib/bots/versionbot.ts
hedss: This bit of code hasn't changed. The `repository.yml` specifics an object for tags now, which means more than once instance of the key can't occur.
The search code searches for all occurrences of the tag name (of which there could be multiple in a commit) and then checks the value for each of those with a valid value ***and breaks out of the loop as soon as it finds one***. This inherently means you'll never get more than one positive result for a commit, regardless of the number of times the tag occurs, ensuring that the final count for all commits is true.

lib/bots/versionbot.ts
lurch: As well as #27 I guess this could also come under #139 ?

hedss: Exactly.

lib/bots/versionbot.ts
```diff
@@ -1693,30 +1696,38 @@ export class VersionBot extends ProcBot {
...
 		let tagResults: MissingTag[] = [];
-		for (let tag of tagDefinitions) {
+		_.each(sanitisedDefs, (tag, name) => {
 			// Lookup the tag in the tag map
-			const tagCount = tagCounts[tag.name];
+			const tagCount = tagCounts[name];
 			let tagsRequired = 0;
 
 			// Convert all/never to numbers, appropriately. As we pre-validated
 			// tags, we don't need to check for 'never' as that's the only
```
lurch: Now seems to be a conflict in this comment ;)

lurch:
Apologies for the post-merge review - I started reviewing it, got side-tracked onto something else, and only just got back to finishing this ;-)

Issue #120, opened 2017-06-01
SyncBot: Message formatting


sqweelygig:
There are a couple of places where message formatting could use improvement, at least:
Front replies don't footer off properly.

sqweelygig:
Also, from Sonya, minimise the visibility of metadata, especially to the public

PR #171, opened 2017-07-03
Format metadata in a more subtle fashion

sqweelygig:
Connects to #120 

Introducing a new hero to my SyncBot reviewer list, Sonya!  I hope she'll be vetting my changes on behalf of the lovely public audience.

front

* Extend metadata encoder to include img and human formats
* Improve metadata extractor be selective upon metadata format
* Give Messengers a message of the day method
* Encode Discourse metadata in a signature style block
* Encode Front reply metadata in a signature style block
* Encode Front comments metadata in a human readable block
* Refactor indicator configuration blocks into objects

Change-Type: minor


lib/services/front.ts
```diff
@@ -135,13 +136,14 @@ export class FrontService extends Messenger implements ServiceListener, ServiceE
...
 				// The specific form that may be emitted
+				const footer = `${Messenger.stringifyMetadata(data, 'img')} ${Messenger.messageOfTheDay()}`;
 				return {
 					endpoint: {
 						method: this.apiHandle.front.message.send,
 					},
 					payload: {
 						author_id: userId,
-						body: `${data.text}<hr/><br/>${Messenger.stringifyMetadata(data, 'plaintext')}`,
+						body: `${data.text}<hr />${footer}`,
```
hedss: Is there meant to be a space in the thematic break tag?

lib/services/front.ts
```diff
@@ -164,31 +166,33 @@ export class FrontService extends Messenger implements ServiceListener, ServiceE
...
 			}
+			const footer = `${Messenger.stringifyMetadata(data, 'img')} ${Messenger.messageOfTheDay()}`;
 			return {
 				endpoint: {
 					method: this.apiHandle.front.message.reply,
 				},
 				payload: {
 					author_id: details.userId,
-					body: `${data.text}<hr/><br/>${Messenger.stringifyMetadata(data, 'plaintext')}`,
+					body: `${data.text}<hr />${footer}`,
```
hedss: Space in thematic break again?

lib/services/messenger-types.ts
```diff
@@ -106,6 +106,12 @@ export interface MessengerEmitResponse extends ServiceEmitResponse {
 	};
 }
 
+export interface PublicityIndicator {
```
hedss: Typedoc comments please!

sqweelygig:
They shouldn't be an issue, and I think are generally considered more standards compliant on tags without content, but are also not a required change; so are removed.

hedss:
@sqweelygig My fault, hadn't realised it was meant to be an empty tag. You could have kept them! (Just a comment to as why without removing them would have been fine)

hedss:
I'm happy with this PR on the explicit proviso:

We discussed yesterday re-architecting the code to ensure that the Front, Flowdock and Discourse services move towards a single-purpose, slimline API Emitter service (as they're intended to be). This means that any Messenger Service specifics will move within the Messenger Service and any sub-components it requires (such as message translation facilities for each dependent service).

It is intended that this should happen after the next couple of major tasks on @sqweelygig's list.

sonyagreen:
We totally talked about this previously but I don't remember - the footer is necessary for some reason, right? It brings the metadata? If I recall correctly there's no reason for the footer to show to the public, it's only for our use.

Any reason not to make it shorter, just "[resin-logo] resin.io"?

sqweelygig:
Yeah the footer hides the metadata, it's for SyncBot's use.  It's either that or a one pixel transparent image which increases your "might be spam" rating; quite a lot from what I remember.
The footer can go down to "[resin-logo] resin.io" - the message after the logo is configured in the environment, not hardcoded.

sonyagreen:
Ah, I see why we wouldn't want to have the transparent image. 

Since the footer isn't necessary for the public, the least offensive/confusing/annoying footer possible would be our goal ... unless we could think of something clever that would be *better* than a gratuitous footer. I can't think of anything (thus the logo/resin.io footer is 👍) but thought I'd drop it here for these smart brains to consider.

sqweelygig:
@sonyagreen the comment after the logo is pulled from the configuration - so is really easy to be [logo] resin.io, and the code as it stands supports this.

sqweelygig:
@sonyagreen with my assurances that the text is fully configurable, could you give this an approve please?

hedss:
@sqweelygig Might be worth re-adding Sonya as a reviewer as it'll kick her notification list.

sqweelygig:
She's given it the thumbs up.  Is getting @sonyagreen to actually click the approve button just process for process' sake?

hedss:
My POV is if you've asked someone for review and they have an opinion on it, they should approve unless it's already been agreed it's not a requirement.

Issue #49, opened 2017-02-22
VersionBot: `status` event not surpressed by `procbots/versionbot/no-checks` label


hedss:
Witnessed here in the logs:
https://github.com/resin-io/resin-image-maker/pull/52
```
22.02.17 16:30:11 [+0000] 2017-02-22T16:30:11.043Z: DEBUG - StatusChangeState: checking version for resin-io/resin-image-maker#52
22.02.17 16:30:11 [+0000] 2017-02-22T16:30:11.499Z: DEBUG - StatusChangeState: No valid 'Change-Type' tag found, failing last commit fo
r resin-io/resin-image-maker#52
22.02.17 16:30:23 [+0000] 2017-02-22T16:30:23.075Z: DEBUG - StatusChangeState: checking version for resin-io/resin-image-maker#52
22.02.17 16:30:23 [+0000] 2017-02-22T16:30:23.433Z: DEBUG - StatusChangeState: No valid 'Change-Type' tag found, failing last commit fo
r resin-io/resin-image-maker#52
22.02.17 16:31:26 [+0000] 2017-02-22T16:31:26.081Z: INFO - Dropping 'CheckForReadyMergeState' as not all trigger labels are present
22.02.17 16:31:33 [+0000] 2017-02-22T16:31:33.420Z: INFO - Dropping 'CheckForReadyMergeState' as not all trigger labels are present
22.02.17 16:31:43 [+0000] 2017-02-22T16:31:43.013Z: INFO - Dropping 'CheckForReadyMergeState' as not all trigger labels are present
22.02.17 16:31:43 [+0000] 2017-02-22T16:31:43.110Z: DEBUG - CheckVersionistCommitStatus: checking version for resin-io/resin-ui#456
22.02.17 16:33:52 [+0000] 2017-02-22T16:33:52.056Z: DEBUG - StatusChangeState: checking version for resin-io/resin-image-maker#52
22.02.17 16:33:52 [+0000] 2017-02-22T16:33:52.208Z: DEBUG - StatusChangeState: No valid 'Change-Type' tag found, failing last commit fo
r resin-io/resin-image-maker#52
```

hedss:
Not sure this is true any longer, need to look.

hedss:
This is actually still happening. @CameronDiver has a seen the issue (see [this](https://www.flowdock.com/app/rulemotion/r-process/threads/qYONSiWocpKn0yYdhIdYUq3m2pC) flowdock thread).

PR #176, opened 2017-07-06
VersionBot: Ensure that the `procbots/versionbot/no-checks` label is correctly honoured

hedss:
Previously status checks could still force VersionBot to make checks on tags in a commit
as there was no way to test the label conditions. This patch ensures that the status
check code itself checks labels and suppresses checks should the labels be present.

Connects-to: #49

Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #49


hedss:
@CameronDiver @pcarranzav This fixes it properly to stop VersionBot checking *anything* (so useful whilst you're still in progress before the work to supress warnings but still do checks has been carried).

lib/bots/versionbot.ts
```diff
@@ -351,6 +350,29 @@ export class VersionBot extends ProcBot {
...
+					},
+					method: this.githubApi.issues.getIssueLabels
+				}).then((labels: GithubApiTypes.IssueLabel[]) => {
+					// If the suppression labels are present, then we remove the PR from the
+					// list of those to check.
+					if (_.find(labels, (label) => label.name === IgnoreLabel)) {
+						this.logger.log(LogLevel.DEBUG,
+							`Dropping '${registration.name}' as suppression labels are all present`);
+						_.remove(prEvents, (labeledEvent) => labeledEvent === prEvent);
+					}
```
pimterry: Not _necessary_, but you can simplify this section, and avoid doing nasty list mutation, if you instead use `Promise.filter` instead of `Promise.map`, and change this whole block to just:

```typescript
return _.every(labels, (label) => label.name !== IgnoreLabel)
```
and then use the resulting filtered array of PR events in the below `then`.

lib/bots/versionbot.ts
```diff
@@ -351,6 +350,29 @@ export class VersionBot extends ProcBot {
...
+				});
+			});
+		}).then(() => {
```
pimterry: If you do the above, you can also actually combine this `then` and the below `Promise.map`, and replace this line with a direct call to map on the existing promise chain:

```typescript
}).map((prEvent) => { ...
```

lib/bots/versionbot.ts
hedss: I have no idea why I didn't use filter in the first place. Good call. :)

Issue #182, opened 2017-07-12
SyncBot: Third-party authorship goes missing when entering Front


sqweelygig:
The user who contributed a forum topic gets lost as it enters Front.  Example here https://app.frontapp.com/open/cnv_68wj2d

sqweelygig:
We're going to have to revisit this one, because testing revealed that the merged fix breaks other things badly.

sqweelygig:
Note to self: This is going to be creating a customer in Front for every forums user, which is not a Bad Thing

PR #183, opened 2017-07-12
SyncBot, ProcBots: Cite sender in Front updates

sqweelygig:
I'm going to be quite explicit in this PR message, because I want to keep the commit history a bit more clean.  I really do not like this.  I've struggled for longer than I wanted to trying to find an alternative place to keep this context.  If any of the reviewers would like a Front instance suitable for random tests they are welcome to ask me.

Connects to #182 

This puts the sender of a message into the data that is displayed on Front. I've tried sender_name and author_id from http://dev.frontapp.com/#send-reply without success.

* Cite message username in visible Front data, specifically `to`.

Change-Type: patch


hedss:
@sqweelygig Can you add Jack and other appropriate reviewers too, please? I think a minimum of two/three is acceptable.

lib/services/front.ts
```diff
@@ -187,6 +187,10 @@ export class FrontService extends Messenger implements ServiceListener, ServiceE
 						archive: false,
 					},
 					subject: details.conversation.subject,
+					// Yes, this is the best place I've found to cite the author of a message for now. SyncBot channels
+					// do not actually use there outgoing triggers anyway, so it doesn't matter *too* much.
```
hedss: `there` -> `their`

sqweelygig:
I've put @sonyagreen on this.  This time it isn't a "no, really, review this", but more of a "FYI, I'm going to be rendering this into a something for Front soon"

lib/services/front.ts
```diff
@@ -187,6 +187,10 @@ export class FrontService extends Messenger implements ServiceListener, ServiceE
...
+					// But it is still wrong.
+					to: [data.toIds.user],
```
hedss: Is this because Front doesn't actually care about the author when processing the message? The API seems to suggest you get the author back correctly on receiving messages.

sqweelygig: To clarify, when you say "API" do you mean "API" or "API Documentation"

hedss: API docs. I haven't actually tried the API (I trust you :) ). So the docs say you can get the author back, but the API doesn't?

sqweelygig: I'll get some screenshots of exactly what I mean.

sqweelygig: front
I don't care too much about whether the codebase or the webhooks or whatever can get the author back.  I care about what the UI displays.

Issue #185, opened 2017-07-13
SyncBot: Sometimes displays multiple comments in Flowdock


sqweelygig:
There a weird bug where a SyncBotted thread sometime displays quite a few copies of a comment in the flow.  However when you click into the thread itself there's only one copy.  Examples:
* https://www.flowdock.com/app/rulemotion/public-s-community/threads/9WGsr77oE7yDnelqkt_F7-mktqQ
* https://www.flowdock.com/app/rulemotion/public-s-community/threads/KlPcWAh2GSTlFyh3dd8wwm26bD4
* https://www.flowdock.com/app/rulemotion/public-s-community/threads/0DlyroZZiEf2Zb6PE7oW2k6zb0g
* https://www.flowdock.com/app/rulemotion/public-s-premium/threads/mTVYkmxnDPHprY9eytrl4zMVZgg

sqweelygig:
![flowdock_multipost](https://user-images.githubusercontent.com/20857369/28157444-7197721c-67ae-11e7-857f-5ec62e78ac2b.png)

sqweelygig:
![flowdock_multipost_b](https://user-images.githubusercontent.com/20857369/28158009-3ca29152-67b0-11e7-9958-65749508b42d.png)
(a thread doesn't agree between web view and app view)

sqweelygig:
Okay, so I'm pretty confident that the immediate issue is down to Flowdock and I'm attempting to open a support case there.
This thread https://www.flowdock.com/app/rulemotion/london_office/threads/w6tIsRldeWfpue9NI5eyVeqbx2A triggers Hubot several times, appears in the API with repetitions, and appears in web inspector with repetitions; despite being created without SyncBot and usually appearing in the UI only once.

sqweelygig:
This was a bug in Flowdock and also now has some defensive coding to capture.

PR #189, opened 2017-07-13
SyncBot, ProcBots: Defend against repeatedly emitted events

sqweelygig:
Connects to #185 

I'd appreciate a little promptness with this one, I've got a deploy I want to bundle tomorrow, and I really want this to be part of it.

* Flowdock adapter now maintains and checks a list of events received

Change-Type: patch


lib/services/flowdock.ts
```diff
@@ -28,6 +28,8 @@ import { ServiceEmitter, ServiceListener } from './service-types';
 
 export class FlowdockService extends Messenger implements ServiceEmitter, ServiceListener, DataHub {
 	private static _serviceName = path.basename(__filename.split('.')[0]);
+	// There are circumstances in which one flowdock event will be received more than once, so track.
+	private postsSynced = new Set<number>();
```
pimterry: Technically, you don't know if you've synced these posts - only that you've added them to a queue to sync hopefully in future. `receivedPostIds`?

Issue #195, opened 2017-07-17
SyncBot: Long messages cause Flowdock error


sqweelygig:
When a Front conversation has a lot of bouncing backwards and forwards due to being an email with previous message quotes SyncBot cries.  It shouldn't.
https://app.frontapp.com/open/cnv_69994d

PR #198, opened 2017-07-17
SyncBot, ProcBots: Ensure long messages are represented in Flowdock.

sqweelygig:
This is tagged as ProcBots since we are ahead of the Great Refactor, so changes to message emitters have a broader than proper impact.
E-mail replies can get rather long, especially when the include the previous conversation, and Flowdock is limited to 8096 characters per message.

* Use the message.text from Front where available.
* Trim the message when posting to Flowdock if required.

Connects-To: #195
---- Autogenerated Waffleboard Connection: Connects to #195


sqweelygig:
This is the last of the urgent fixes on the to-do list at the moment, so if it gets an approval today then I can cut a version for deployment.

lib/services/flowdock.ts
```diff
@@ -120,8 +120,16 @@ export class FlowdockService extends Messenger implements ServiceEmitter, Servic
...
 	public makeSpecific = (data: TransmitContext): Promise<FlowdockEmitContext> => {
-		// Build a string for the title, if appropriate.
+		// Calculate details such as title, footer and snipped message, if appropriate.
+		const lengthLimit = 8096;
 		const titleText = data.toIds.thread ? '' : data.title + '\n--\n';
+		const footerText = Messenger.stringifyMetadata(data, 'emoji');
+		let trimText = '\n\n`... about xx% shown.`';
+		const trimmedText = data.text.substr(0, lengthLimit - titleText.length - trimText.length - footerText.length);
+		trimText = trimText.replace('xx', Math.floor((100*trimmedText.length)/data.text.length).toString(10));
+		const bodyText = titleText.length + data.text.length + footerText.length < lengthLimit
```
hedss: I'd personally add parentheses here, purely to make it obvious, but that's a comment, not a requirement.

lib/services/flowdock.ts
```diff
@@ -137,7 +145,7 @@ export class FlowdockService extends Messenger implements ServiceEmitter, Servic
 				},
 				payload: {
 					// The concatenated string, of various data nuggets, to emit
-					content: titleText + data.text + Messenger.stringifyMetadata(data, 'emoji'),
+					content: titleText + bodyText + footerText,
```
hedss: Probably worth just doing string literal here instead of addition.

lib/services/flowdock.ts
```diff
@@ -120,8 +120,16 @@ export class FlowdockService extends Messenger implements ServiceEmitter, Servic
 	 * @returns     Promise that resolves to the emit suitable form.
 	 */
 	public makeSpecific = (data: TransmitContext): Promise<FlowdockEmitContext> => {
-		// Build a string for the title, if appropriate.
+		// Calculate details such as title, footer and snipped message, if appropriate.
+		const lengthLimit = 8096;
```
CameronDiver: This might not make sense in this context, but maybe it's worth exporting this to a global constant, and also checking the env so it can be changed without a redeploy:
```typescript
const LENGTH_LIMIT = process.env.LENGTH_LIMIT || 8096

export class FlowdockService ...

File: lib/services/flowdock.ts [hunk 4]
```diff
@@ -120,8 +120,16 @@ export class FlowdockService extends Messenger implements ServiceEmitter, Servic
...
 		const titleText = data.toIds.thread ? '' : data.title + '\n--\n';
+		const footerText = Messenger.stringifyMetadata(data, 'emoji');
+		let trimText = '\n\n`... about xx% shown.`';
+		const trimmedText = data.text.substr(0, lengthLimit - titleText.length - trimText.length - footerText.length);
+		trimText = trimText.replace('xx', Math.floor((100*trimmedText.length)/data.text.length).toString(10));
```
CameronDiver: It'd be nice if we can generate this only if needed.

PR #206, opened 2017-07-18
Withdraw Front source citation bodge

sqweelygig:
connects to #182 

So, integration testing revealed a bug in the Front citation code, which was a bodge anyway; so it's going to have to be reconsidered.

* Remove Front citation bodge, because it breaks emailed topics.
* Analyse Front messages using full body, to pick up img encoded metadata.
* Continue to process Front events with just the text, no html.

Change-Type: patch


sqweelygig:
`syncbot-alpha` branch has now received a batch of testing within minikube.

Issue #64, opened 2017-03-14
VersionBot: Nice to have: Block normal merges


pimterry:
[At least once](https://github.com/resin-io/resin-ui/pull/523) I've forgotten about versionbot, and merged PRs manually, leading to an annoying manual version update process. It'd be nice to versionbot took total responsibility of PRs (except for no-checks), and blocked normal merges completely, so that I can't accidentally usurp it and screw things up.

lurch:
AFAIK there'd be no way for VersionBot to completely block normal merges, but at least on Etcher (which isn't using VersionBot yet) the repo is set up so that only the project-maintainer is allowed to merge PRs.

pimterry:
Still allows things to go wrong if you're the maintainer :smile:. The UI is set up to block merges if there's a failing status check, and Versionist does fail and block merges there when there's no change-type footer, it's just that it then unblocks them once there is. If they simply stayed blocked that'd solve this. Might have some other UX implications though (might mean that status markers on commits are always ❌?)

lurch:
OTOH, preventing any "manual override" of VersionBot by the project-maintainer if / when things go wrong, might cause bigger problems? \*shrug\*

hedss:
Several issues have now been seen in reverse, where people have hit the 'Merge' button without thinking and got `master` out of sync. This is a problem, and one solution is to hide the button under another status check that only resolves when the PR is approved. We actually want to remove the `procbots/versionbot/ready-to-merge` label (see #72).

hedss:
![](https://github.trello.services/images/mini-trello-icon.png) [VersionBot: Nice to have: Give feedback on the changelog lines to be added in advance](https://trello.com/c/oc7rzTDh/9-versionbot-nice-to-have-give-feedback-on-the-changelog-lines-to-be-added-in-advance)

PR #210, opened 2017-07-19
VersionBot: Ensure that manual merges cannot occur by adding a new status check

hedss:
This functionality stops the merging of PRs by anybody other than VersionBot (although
admin users can override this). This prevents accidental use of the 'Merge' button
on Github, instead of using the `procbots/versionbot/ready-to-merge' label.

Connects-to: #64

Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #64


lib/bots/versionbot.ts
```diff
@@ -795,7 +842,7 @@ export class VersionBot extends ProcBot {
 			let description = 'Missing or forbidden tags in commits, see `repository.yml`';
```
lurch: Should this be moved to a constant further up the file, along with the other ContextStatusMessages?
(unless there's some good reason for this being `let` rather than `const`?)

lib/bots/versionbot.ts
```diff
@@ -1369,11 +1424,14 @@ export class VersionBot extends ProcBot {
 
 	/**
 	 * Retrieve all protected branch status requirements, and determine the state for each.
+	 * Status checks can be filtered to ignore the results of given contexts, if required.
```
lurch: Shouldn't the `ignore` here be `include/exclude` ?

lib/bots/versionbot.ts
```diff
@@ -1429,13 +1488,22 @@ export class VersionBot extends ProcBot {
 			_.each(protectedContexts, (proContext) => {
 				// We go through every status and see if the context prefixes the context
 				// of the status.
+				// We filter the VersionBot reviews.
```
lurch: nit: there appears to be nothing in the filtering below hardcoded to `VersionBot` ?

lib/bots/versionbot.ts
```diff
@@ -1429,13 +1488,22 @@ export class VersionBot extends ProcBot {
...
-						statusResults.push({
-							name: status.context,
-							state: statusLUT[status.state]
-						});
+						let includeContext = true;
+						// If there's a filter, determine if the result should be included or excluded.
+						if (filter) {
+							includeContext = filter.includeContexts ?
+								_.find(filter.contexts, (context) => context === status.context) !== undefined :
+								!_.find(filter.contexts, (context) => context === status.context);
```
lurch: nit: Given that both branches of the ternary use `_.find(filter.contexts, (context) => context === status.context)`, I guess this code could be compacted slightly with the use of another variable? e.g.
```
const contextIsFiltered = _.find(filter.contexts, (context) => context === status.context);
includeContext = filter.includeContexts ? contextIsFiltered  !== undefined : !contextIsFiltered;
```
(and I'm not a TS programmer, but maybe that could be compacted further too)

lib/bots/versionbot.ts
hedss: Good spot.

lib/bots/versionbot.ts
```diff
@@ -186,6 +186,19 @@ enum StatusChecks {
...
 
+/**
+ * Allows the filtering of results from the `checkStatuses` method.
+ * Depending on the `includedContexts` flag, results returned from the method:
+ *   * Only include results from contexts listed in the `contexts` property (`includedContexts` is `true`)
+ *   * Only include results from contexts not listed in the `contexts` property (`includedContexts` is `false`)
+ */
+interface StatusFilter {
+	/** Whether this filter includes tags that should be included or excluded. */
+	includeContexts: boolean;
```
hedss: Fixed.

lib/bots/versionbot.ts
```diff
@@ -723,16 +757,29 @@ export class VersionBot extends ProcBot {
...
 		const author = prEvent.sender.login;
+		const prAction = event.cookedEvent.data.action;
 		let committer = author;
 		let lastCommit: GithubApiTypes.Commit;
 		let botConfig: VersionBotConfiguration;
 
 		// Only for opened or synced actions.
-		if ((event.cookedEvent.data.action !== 'opened') && (event.cookedEvent.data.action !== 'synchronize') &&
-			(event.cookedEvent.data.action !== 'labeled')) {
+		if ((prAction !== 'opened') && (prAction !== 'synchronize') && (prAction !== 'labeled')) {
```
hedss: Could switch to using Lodash, I guess.

Issue #73, opened 2017-03-28
VersionBot: Allow authors of a PR to also start merge should there be a set of maintainers


hedss:
The reasoning behind #72 came down to ensuring that the pipeline wasn't delayed if maintainers weren't able to get through PRs swiftly. This is compounded if there's a configuration file that only allows certain engineers (the maintainers of a repository) to start the merge.

The solution is to allow either authorised maintainers **or** the original author of the PR to apply the `procbots/versionbot/ready-to-merge` label.

Should be a quick fix.

lurch:
Should this also be a feature that's configurable on a per-repo basis using `procbots.yml` ?

I guess this feature needs to be used with care on any public repos where Joe Bloggs can submit a PR to a resin.io repo - we obviously wouldn't want VersionBot automatically merging random PRs from the public just because the original author happened to add a `procbots/versionbot/ready-to-merge` label.

lekkas:
@lurch that's a fair point. For private repos access control is performed by github essentially so it should be fine to act on a  `
procbots/versionbot/ready-to-merge` label set by the PR Author.

For public repos, setting the label will only have an effect if the user who adds it is listed in the `maintainers` property of `.procbots.yml`

Does this make sense?

lekkas:
Connects to https://github.com/resin-io/hq/pull/730

lurch:
...but then resin.io employees (who aren't listed as a maintainer in any repos) might get confused why they _can_ apply the `procbots/versionbot/ready-to-merge` label on a private repo, but _can't_ apply the label on a public repo? (when you're jumping between multiple repos, it's not immediately obvious which is which unless you remember to double check).

I hesitate to suggest it :confounded: but does VersionBot need to be able to lookup a user's organisation membership, so that it can tell which users are (current) resin.io employees? Or does that not give fine-grained enough details? \*shrug\*

alexandrosm:
@lekkas confused. Are we going back on author being able to merge if
approved?

On Mar 30, 2017 5:01 AM, "Andrew Scheller" <notifications@github.com> wrote:

...but then resin.io employees (who aren't listed as a maintainer in any
repos) might get confused why they *can* apply the
procbots/versionbot/ready-to-merge label on a private repo, but *can't*
apply the label on a public repo? (when you're jumping between multiple
repos, it's not immediately obvious which is which unless you remember to
double check).

I hesitate to suggest it 😖 but does VersionBot need to be able to lookup a
[...]

PR #214, opened 2017-07-21
VersionBot: Ensure that authors of a PR are allowed to merge it even if a maintainers list is present

hedss:
This ensures that an author can also take responsibility for a PRs merging without
the need for a maintainer to do so.

Connects-to: #73

Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #73


lib/bots/versionbot.ts
```diff
@@ -1682,13 +1682,17 @@ export class VersionBot extends ProcBot {
...
 		if (maintainers) {
+			// A user is a 'special' maintainer, if all reviews have been approved.
+			// Essentially this provides a mechanism for ensuring responsibility.
+			maintainers.push(event.pull_request.user.login);
+
 			// Get the user who added the label.
 			if (!_.includes(maintainers, event.sender.login)) {
 				let errorMessage = `The \`${MergeLabel}\` label was not added by an authorised ` +
-					'maintainer. Authorised maintainers are:\n';
+					'maintainer or by the PR author. Authorised mergers are:\n';
```
CameronDiver: It's kind of useless but it might be nice to have (which I why I'm adding as a comment) but you could say something like `.... The ${maintainers.length} authorised mergers are:\n`

PR #218, opened 2017-07-24
Ensure checks/updates are always on the head regardless of owner; move to a more flexible Service usage system

hedss:
Normally these would be two separate issues, but they were originally part of the move to allow forked contributor and insta-merge to occur, hence one was built on top of the other.
---- Autogenerated Waffleboard Connection: Connects to #215
---- Autogenerated Waffleboard Connection: Connects to #216


lib/bots/versionbot.ts
```diff
@@ -239,21 +239,23 @@ const StatusAutoMerge = {
 export class VersionBot extends ProcBot {
 	/** Github ServiceListener name. */
 	private githubListenerName: string;
-	/** Github ServiceEmitter name. */
+	/** Github ServiceListener name. */
 	private githubEmitterName: string;
```
CameronDiver: What's going on here? You have the comment saying that it's a `Listener`, but the varname is still `Emitter`.

lib/services/github.ts
```diff
@@ -488,42 +504,51 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
...
+
+				// Request new token.
+				const tokenOpts: any = {
+					headers: {
+						Accept: 'application/vnd.github.machine-man-preview+json',
+						Authorization: `Bearer ${jwToken}`,
+						'User-Agent': 'request'
+					},
+					json: true,
+					method: 'POST',
```
CameronDiver: I believe that you don't need this if you use the `request.post` method.

lib/utils/environment.ts
```diff
@@ -50,7 +52,8 @@ export function BuildCommand(command: string, args?: string[], options?: Command
 	if (options) {
 		builtOptions = {
 			cwd: options.cwd,
-			retries: options.retries
+			retries: options.retries,
+			delay: options.retries
```
CameronDiver: This should be `options.delay` I guess?

hedss: Good catch!

lib/bots/versionbot.ts
hedss: C+P fatigue. :)

lib/services/github.ts
hedss: Good catch.

lib/utils/environment.ts
```diff
@@ -75,6 +78,7 @@ export function BuildCommand(command: string, args?: string[], options?: Command
  */
 export function ExecuteCommand(command: Command): Promise<{}> {
 	let tries = ((command.options || {}).retries || 0) + 1;
+	let delay = (command.options || {}).delay || 1000;
```
CameronDiver: I've seen this pattern in quite a few places in the codebase, it might be better to (eventually) migrate to [_.get()](https://lodash.com/docs/4.17.4#get), which would make these two cases:
```typescript
let tries = (_.get(command.options, 'retries') || 0) + 1
let delay = _.get(command.options, 'delay') || 1000
```
It doesn't really add much for these smaller cases, except perhaps slightly more readable, but there is other places I could see it being quite useful.

lib/services/github.ts
```diff
@@ -488,42 +504,50 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
...
-		return request.get(installationsOpts).then((installations) => {
-			// Get the URL for the token.
-			const tokenUrl = installations[0].access_tokens_url;
+		// Authentication depends on App or PAT.
+		let tokenPromise: Promise<AuthToken>;
 
-			// Request new token.
-			const tokenOpts: any = {
+		if (this.appId) {
+		// Initialise JWTs
```
CameronDiver: This should be indented one.

lib/services/github.ts
```diff
@@ -488,42 +504,50 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
...
-				method: 'POST',
-				url: tokenUrl
+				url: 'https://api.github.com/app/installations'
 			};
 
-			return request.post(tokenOpts);
-		}).then((tokenDetails) => {
+			tokenPromise = request.get(appOpts).then((apps) => {
+				// Get the URL for the token.
+				const tokenUrl = apps[0].access_tokens_url;
```
CameronDiver: This could throw if no applications are returned, is that possible?

lib/services/github-types.d.ts
```diff
@@ -18,38 +18,36 @@ import * as Promise from 'bluebird';
...
 import TypedError = require('typed-error');
 import { ConfigurationLocation, ProcBotConfiguration } from '../framework/procbot-types';
-import { ServiceAPIHandle, ServiceEmitContext, ServiceEvent, ServiceListenerMethod, ServiceRegistration,
-	ServiceType } from './service-types';
+import { ServiceAPIHandle, ServiceBase, ServiceEmitContext, ServiceEvent,
+	ServiceListenerMethod, ServiceRegistration, ServiceType } from './service-types';
 
 /** The Github service can be used as either an Integration or a User. */
-type GithubLoginType = 'integration' | 'user';
+type GithubLoginType = 'app' | 'pat';
```
CameronDiver: Does this make sense to be an enum or does it need to referenced by string?

lib/framework/procbot.ts
```diff
@@ -172,32 +176,39 @@ export class ProcBot {
...
 		// whether this is an issue or not
-		return Promise.map(this.emitters, (emitter) => {
-			return emitter.sendData(data)
-			.then((result) => { results.push(result); })
-			.catch((error) => { results.push(error); });
-		}).return(results);
+		this.emitters.forEach((emitterEntry) => {
+			const emitter = <ServiceEmitter>emitterEntry.instance;
+			const promise = emitter.sendData(data)
+				.then((result) => { results.push(result); })
```
CameronDiver: Did you mean to change the indentation here?

lib/framework/procbot.ts
```diff
@@ -108,58 +118,52 @@ export class ProcBot {
...
-		const service = this.getService(name);
-		let listener;
-
-		if (service && !_.find(this.listeners, [ 'serviceName', name ])) {
-			listener = service.createServiceListener(data);
-			this.listeners.push(listener);
-		}
-
-		return listener;
+		return <ServiceListener>this.addService('listener', name, data);
```
CameronDiver: Completely preference-based, but I think this could be more readable as
```typescript
return this.addService('listener', name, data) as ServiceListener
```
and again lower down.

lib/utils/environment.ts
CameronDiver: In fact, we could make it even smaller:
```typescript
let tries = _.get(command.options, 'retries', 0) + 1
let delay = _.get(command.options, 'delay', 1000)

Comment (pimterry) [inline lib/utils/environment.ts]: Alternatively, `BuildCommand` could provide a default `.options` object (`{}` would help, something like `{ retries: 0, delay: 1000, ... }` with `_.defaults` would be even better), and simplify all this downstream code completely.

File: lib/services/github.ts (hunk 6 above)
Comment (hedss) [inline lib/services/github.ts]: Nope, you'll always get an empty array. However, the `apps[0]` will throw if it's empty, and we rely on that to stop the ProcBot run, as it shows that there's something wrong with the App details (essentially it's an invalid VersionBot instance).

File: lib/framework/procbot.ts (hunk 8 above)
Comment (hedss) [inline lib/framework/procbot.ts]: Yep, it makes it more obvious (to me, anyway!) that they belong to the `const` declaration on the line above.

File: lib/services/github-types.d.ts [hunk 10]
```diff
@@ -22,24 +22,22 @@ import { ServiceAPIHandle, ServiceEmitContext, ServiceEvent, ServiceListenerMeth
...
 
 /** Allows a service to login to Github as a user, using username and Personal Access Token. */
-export interface GithubUser {
+export interface GithubPAT {
 	/** The PAT for the user to login as. */
 	pat: string;
-	/** The username of the user to login as. */
-	user: string;
 	/** Type of service ('listener' or 'emitter'). */
 	type: GithubLoginType;
```
pimterry: This comment (and the equivalent one for `GithubApp`) is now wrong I think - it looks like this is now 'app' | 'pat'

In addition, it's not even that: for each type it's _always_ one of those: `type: 'pat'` here and `type: 'app'` for `GithubApp`.

lib/services/github.ts
```diff
@@ -101,14 +115,16 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
...
-		if (constObj.loginType.type !== 'integration') {
-			throw new Error('Do not yet support non-Integration type clients');
+		if (constObj.authentication.type === 'app') {
+			const appLogin = <GithubApp>constObj.authentication;
+
+			// Set the App ID
+			this.appId = appLogin.appId;
+			this.pem = appLogin.pem;
+		} else {
+			const userLogin = <GithubPAT>constObj.authentication;
```
pimterry: If you implement the `type: 'pat'` change in my comment above, you shouldn't need the cast here or in the previous block: TS should be able to know what type `authentication` has based on the `if` test above.

lib/services/github.ts
```diff
@@ -488,42 +504,51 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
...
 			};
 
-			return request.post(tokenOpts);
-		}).then((tokenDetails) => {
+			tokenPromise = request.get(appOpts).then((apps) => {
+				// Get the URL for the token.
+				const tokenUrl = apps[0].access_tokens_url;
+
+				// Request new token.
+				const tokenOpts: any = {
```
pimterry: Is there a reason you want this as `any`? If you change this code to inline the options in the call (`return request.post({ headers: ... })`) then you don't need to define the type explicitly, and you'll get type checking from whatever definitions we have available for `request.post` to make sure your params are all valid.

lib/framework/procbot.ts
pimterry: Even better, rather than casting at all, you should be able to write an overload for `addService` that tells TS that if the first param is `listener`, it returns a `ServiceListener`, and if the first param is `emitter` it returns a `ServiceEmitter`. That'll drop this cast entirely, and the other one below, and any future ones.

lib/framework/procbot.ts
```diff
@@ -36,8 +46,8 @@ const exec: (command: string, options?: any) => Promise<{}> = Promise.promisify(
 export class ProcBot {
 	protected _botname: string;
 	protected logger = new Logger();
-	private emitters: ServiceEmitter[] = [];
-	private listeners: ServiceListener[] = [];
+	private listeners: Map<string, ServiceMap> = new Map<string, ServiceMap>();
+	private emitters: Map<string, ServiceMap> = new Map<string, ServiceMap>();
```
pimterry: Could we make `ServiceMap` generic in its `.instance` type, so that these maps are known to only contain emitters or listeners up front, and we don't need to cast the results later?

lib/framework/procbot.ts
```diff
@@ -172,32 +176,39 @@ export class ProcBot {
...
-		}).return(results);
+		this.emitters.forEach((emitterEntry) => {
+			const emitter = <ServiceEmitter>emitterEntry.instance;
+			const promise = emitter.sendData(data)
+				.then((result) => { results.push(result); })
+				.catch((error) => { results.push(error); });
+			emitterPromises.push(promise);
+		});
+
+		return Promise.all(emitterPromises).return(results);
```
pimterry: I think you can drop the `emitterPromises` and `results` arrays completely, and simplify this whole function to:

```typescript
return Promise.all(this.emitters.map((emitterEntry) => {
  const emitter = <ServiceEmitter>emitterEntry.instance // you can remove this too if we make ServiceMap generic
  return emitter.sendData(data).catch((err) => err)
}));
```

lib/utils/environment.ts
hedss: Am going with the `_.defaults` suggestion, a good idea.

lib/services/github-types.d.ts
hedss: Is now an enum.

lib/framework/procbot.ts
CameronDiver: @pimterry that's pretty interesting, do you have a link showing details? (Or is it literally done via function signatures?)

pimterry: @CameronDiver demo with enums [here](https://www.typescriptlang.org/play/#src=enum%20ServiceType%20%7B%0A%20%20%20%20Listener%20%3D%200%2C%0A%20%20%20%20Emitter%20%3D%201%0A%7D%20%0Atype%20ServiceEmitter%20%3D%20%7B%7D%0Atype%20ServiceListener%20%3D%20%7B%7D%0Atype%20ServiceConstructor%20%3D%20%7B%7D%0A%0Afunction%20addService(type%3A%20ServiceType.Listener%2C%20name%3A%20string)%3A%20ServiceListener%20%7C%20void%0Afunction%20addService(type%3A%20ServiceType.Emitter%2C%20name%3A%20string)%3A%20ServiceEmitter%20%7C%20void%0Afunction%20addService(type%3A%20ServiceType%2C%20name%3A%20string)%3A%20ServiceListener%20%7C%20ServiceEmitter%20%7C%20void%0A%7B%0A%0A%7D%0A%0A%2F%2F%20Mouseover%20vars%20to%20see%20the%20inferred%20types%0Avar%20a%20%3D%20addService(ServiceType.Listener%2C%20'a')%3B%0Avar%20e%20%3D%20addService(ServiceType.Emitter%2C%20'b')%3B).

You can do the same thing with string literals or anything else TS can match at a type level. You just write a series of function signatures together, it uses the first one it finds that matches in the order they're defined, and the last one has to have a broad enough signature to encompass all the possible previous types, and is what the actual function body has to match.

pimterry: If you really wanna go wild, here's a half-built totally type-safe promisify implementation I was working on a little while ago, which uses this _heavily_: https://github.com/pimterry/typed-promisify-all/blob/master/index.ts. Lets you match functions to make it impossible to accidentally promisify something that doesn't actually take a callback.

Kind of stalled because I was really just getting distracted while trying to solve promisifyAll, and I'm actually not sure that bit is properly doable :cry:

CameronDiver: That's really nice actually, takes away the verbosity but keeps this clear as well. Thanks for the demo :+1:

lib/framework/procbot.ts
```diff
@@ -234,4 +235,43 @@ export class ProcBot {
...
+	 * @param data  ServiceBase extended constructor, giving an optional handle and constructor data.
+	 * @return      A ServiceListener, ServiceEmitter or void value.
+	 */
+	private addService(type: ServiceType, name: string, data?: ServiceConstructor):
+	ServiceListener | ServiceEmitter | void {
+		// If there's no handle in the constructor data, the handle is the name.
+		let handle = data ? data.handle : name;
+		if (!handle) {
+			handle = name;
+		}
```
CameronDiver: These four lines can be replaced by 
```typescript
let handle = _.get(data, 'handle', name)

File: lib/framework/procbot.ts [hunk 16]
```diff
@@ -234,4 +235,43 @@ export class ProcBot {
...
+
+		// Attempt to retrieve the named service.
+		const service = this.getService(name);
+		const serviceMap = (type === ServiceType.Listener) ? this.listeners.get(handle) : this.emitters.get(handle);
```
CameronDiver: Completely optional suggestion, the `type === [sometype]` pattern is used a lot, it may be worth adding some functions such as `isServiceListener()` etc.

CameronDiver: It also has the advantage of if the type changes at all, you can change the testing function rather than a big search-and-replace.

lib/services/github.ts
```diff
@@ -48,12 +48,25 @@ interface LabelDetails {
 	};
 }
 
+/** Github Authentication Token. */
+interface AuthToken {
+	/** Token itself. */
+	token: string;
+	/** When the token expires. */
+	expires_at?: string;
```
CameronDiver: Is this snake_case because it comes from the GH api?

lib/utils/environment.ts
```diff
@@ -74,7 +75,8 @@ export function BuildCommand(command: string, args?: string[], options?: Command
  * @throws        An error containing a message from the failed command's `stderr` output.
  */
 export function ExecuteCommand(command: Command): Promise<{}> {
-	let tries = ((command.options || {}).retries || 0) + 1;
+	let tries = (command.options.retries || 0) + 1;
+	let delay = command.options.delay || 1000;
```
CameronDiver: Not sure if I mentioned this in the last comments, but you can use `_.get` here.

lib/services/github.ts
hedss: Yes

lib/utils/environment.ts
hedss: I could. But it'll be doing exactly the same thing with more work. :) This is where I feel it's style over substance, really (then again, if this was `((thing || {}).other || {}).thing || {}`, I'd use it as it were shorter.

lib/framework/procbot.ts
hedss: Done.

lib/framework/procbot.ts
hedss: There's an issue with this though. We check types before we create Service instances. Which means we'd need an implementation in the `ProcBot` class to test. However, there's a couple of other tests in the Service themselves, which won't be touching the `ProcBot` class, so we'd need to define them in the ServiceListener/ServiceEmitter themselves. So we'd end up with two places we'd need to change them anyway.

I want to have a think about this; so, I'm not going to do it now. But it's a very valid suggestion.

Issue #227, opened 2017-07-27
VersionBot: `procbots/versionbot/no-checks` label gets blocked by AutoMerge


hedss:
Whilst using the `no-checks` label now works, it still gets blocked because the other VB requirements need a working status change to allow a merge.

So, we need to see the label and set Versionist/Reviews/etc. checks to pass and allow a manual merge to go ahead (adding the `ready-to-merge` label will simply not be honoured as the `no-checks` label suppresses it).

lurch:
So, AFAICT this feature is desired so that a 'tiny' PR (e.g. https://github.com/resin-io/resin-git/pull/54 ) can be merged without causing versionist to do a 'noisy' full version bump. IMHO rather than changing the semantics around the `no-checks` and `ready-to-merge` labels, I still think the `Change-Type: none` is a better solution to this situation. It feels to me like a more "generic" solution, as it only turns off the version-bumping, and doesn't turn off the other checks e.g. reviewers and required-footer-tags. (but of course there may well be other aspects to this that I'm missing!)

hedss:
@lurch There's been several meetings about this this week, and because of other factors and long-term plans, the consensus is to leave as is currently.
Latest arch call: https://drive.google.com/open?id=0B0NS-URBofBLbGQwRlBJZFBJdE0

hedss:
@lurch I think to be clearer, my original comment here is confusing. What I'm saying is that in the `no-check` case:

* `Versionist` status is automatically passed
* `AutoMerge` status is automatcally passed (to allow manual merging)
* Reviewer checks would still occur, though, because these are *not* related to versionbumping or the merging of VersionBot

Should have been clearer with what I meant by 'set Versionist/Reviews/etc. checks to pass'.

PR #228, opened 2017-07-28
Ensure the `procbots/versionbot/no-checks` label correctly allows manual merging

hedss:
This commit ensures that when the `no-checks` label is applied, the `AutoMerge` and
`Versionist` statuses are automatically successful. It will *not* pass the `Reviewers`
status, however, which still needs to be valid for merging.

Additionally, a couple of bugs have been fixed to allow a move to TypeScript 2.4.x,
mostly around the use of the `Worker` class' generic type usage.

Connects-to: #227

Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #227


lib/framework/procbot.ts
```diff
@@ -136,7 +136,7 @@ export class ProcBot {
 		const service = this.getService(name);
 		let emitter;
 
-		if (service && !_.find(this.emitters, [ 'serviceName', name ])) {
+		if (service && !_.find(this.emitters, (instance) => instance.serviceName === name)) {
```
CameronDiver: This can be:
```typescript
if (service && !_.find(this.emitters, { 'serviceName': name }))

File: lib/framework/procbot.ts [hunk 2]
```diff
@@ -117,7 +117,7 @@ export class ProcBot {
 		const service = this.getService(name);
 		let listener;
 
-		if (service && !_.find(this.listeners, [ 'serviceName', name ])) {
+		if (service && !_.find(this.listeners, (instance) => instance.serviceName === name)) {
```
CameronDiver: Same as below (or above?), this can be shortened.

lib/bots/versionbot.ts
```diff
@@ -448,6 +458,50 @@ export class VersionBot extends ProcBot {
...
+					context: StatusVersionist.Context,
+					description: StatusVersionist.NoVersioning,
+					owner,
+					repo,
+					sha: head.sha,
+					state: 'success'
+				},
+				method: this.githubApi.repos.createStatus,
+			})
+		]).return();
```
CameronDiver: This is a little confusing, you have a `Promise.join`, but no handling function. Should this not be a `Promise.all`?

hedss: Promise.join() is far more performant on small numbers of promises: http://bluebirdjs.com/docs/api/promise.join.html hence that over `.all`. Don't need a handler function as I've no interest in the results.

lib/framework/procbot.ts
hedss: Good catch!

lib/bots/versionbot.ts
pimterry: Personally, I usually use `.all`, because it's clearer, but also more flexible and easy to use, and I worry about that more than the performance difference, which don't make much difference in a project like this.

As a case in point, this code is silently broken :smiley:. `.join` always knows the exact number of promises it wants, because it doesn't take a dynamically sized array: it takes a fixed number of parameters. With `.all` you can pass an array, or a promise that resolves to an array, and get sensible behaviour, but if you pass it an array to `.join` (as you're doing here) then it doesn't examine it, and instead treats it as a resolved final value and silently doesn't wait on promises within at all.

This code isn't waiting for these operations, and doesn't handle errors from them.

More clearly:

```typescript
// This prints 'done' instantly, plus a warning about unhandled errors
Promise.join([ Promise.delay(10000), Promise.reject(new Error('oh no!')) ])
.then(() => console.log('done'))
```

hedss: Hah, and it bit me. :) Altered.

lib/bots/versionbot.ts
```diff
@@ -767,12 +821,19 @@ export class VersionBot extends ProcBot {
...
 		const prAction = event.cookedEvent.data.action;
+		const prLabel: GithubApiTypes.Label = event.cookedEvent.data.label;
 		let committer = author;
 		let lastCommit: GithubApiTypes.Commit;
 		let botConfig: VersionBotConfiguration;
 
-		// Only for opened or synced actions.
-		if (!_.find(['opened', 'synchronize', 'labeled'], (action) => action === prAction)) {
+		// Only for opened, synced or labeling actions.
+		if (!_.find(['opened', 'synchronize', 'labeled', 'unlabeled'], (action) => action === prAction)) {
```
lurch: I don't have much lodash knowledge, but couldn't `_.includes` (or `_.some`) be used instead of `_.find` ?

Issue #230, opened 2017-07-31
VersionBot: Suffix PR number to Changelog entries


hedss:
This will allow NotifyBot to search for released PRs/Issues when a component is deployed, which in turns allows it to notify users of releases.

PR #231, opened 2017-07-31
VersionBot: Append PR number of PR being merged to the `CHANGELOG.md` entry for it

hedss:
This allows tracing of each feature to its PR, and importantly will allow NotifyBot to
do the same to get back to Front conversations for informing customers of feature/fix
releases.

Connects-to: #230 

Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #230


lib/bots/versionbot.ts
```diff
@@ -33,6 +33,7 @@ import { AlertLevel, LogLevel } from '../utils/logger';
 // Exec technically has a binding because of it's Node typings, but promisify doesn't include
 // the optional object (we need for CWD). So we need to special case it.
 const fsReadFile: (filename: string, options?: any) => Promise<Buffer | string> = Promise.promisify(FS.readFile);
+const fsWriteFile: (path: string, contents: string) => Promise<{}> = Promise.promisify(FS.writeFile);
```
CameronDiver: Is it worth using `mz` for these functions? They're already promisified and types are provided.

lib/bots/versionbot.ts
```diff
@@ -1266,6 +1270,28 @@ export class VersionBot extends ProcBot {
...
 				versionData.version = match[1];
 				versionData.files = moddedFiles;
+
+				// Now we have to add the PR number and URL to every log entry we've just added for
+				// the new version. This ensures that NotifyBot can apply release notes later on, if
+				// required.
+				const versions = contents.split('## ');
+
+				// Now find the right version entry.
+				for (let index = 0; index < versions.length; index += 1) {
```
CameronDiver: Given that you only use `index` to retrieve the elements of the array, this should probably be a `for(let version of versions)`

lib/bots/versionbot.ts
hedss: Having had a look, I'm going to stick with what I have atm due to the return of native Promises via `mz` and the mixing with Bluebird. I'd like to investigate what potential interactions should look like correctly. But it does look cleaner. :)

lib/bots/versionbot.ts
hedss: Done.

CameronDiver: Ah yes, didn't notice that. Ignore me :)

lekkas:
👍  all the pieces coming together :)

Issue #238, opened 2017-08-03
ProcBots, VersionBot: Github Service needs to always check label array


hedss:
Else actions will occur that shouldn't.

PR #239, opened 2017-08-03
ProcBots, VersionBot: Ensure Github ServiceListener always checks labels; VersionBot does not act on `flow/in-progress`

hedss:
This patches broken behaviour, denoted in the title.

Change-type: patch


hedss:
Connects-to: #238

lib/bots/versionbot.ts
```diff
@@ -856,7 +856,7 @@ export class VersionBot extends ProcBot {
 
 		// If there's an unlabeling event, but it wasn't the IgnoreLabel being removed, then we
```
lurch: Should this comment have been updated to match the code? Otherwise a future reader won't know if the comment is incorrect or the code is incorrect! :-/

hedss: Will update in future PR.

Issue #70, opened 2017-03-21
NotifyBot: Does not yet exist


hedss:

hedss:
https://github.com/resin-io/process/issues/6

hedss:
![](https://github.trello.services/images/mini-trello-icon.png) [NotifyBot](https://trello.com/c/l6YUcaZ8/3-notifybot)

Issue #85, opened 2017-05-04
NotifyBot: Add comment on PRs that have made it to staging/production


hedss:
See this conversation: 
https://www.flowdock.com/app/rulemotion/resin-tech/threads/eV4Mvt8VnBXWv9Q_HQJP75SlX6p

Essentially it'd be useful to update PRs when they've made it onto staging/production so that PRs/issues aren't left in limbo.

PR #243, opened 2017-08-10
NotifyBot: Implement NotifyBot, a bot for tracking PR issues and commenting on them come deployment

hedss:
* Triggered on commit of Keyframe change to a relevant repo, diffs previous Keyframe to find versions
* Altered versions of a component are tracked, and PRs for all changed versions are followed
* For every PR, connected issues are tracked, as well as linked HQ issues, and any Front conversation for the issue is gathered
* For every PR, a comment is made addressing the author, informing them of the version deployed that the PR is in
* For every Front topic found in an issue (or associated HQ issue), post a comment on all Front conversations that the feature is deployed

This includes work that is being carried out for refactoring the MessengerService, as such the
`front-types.d.ts`, `front.ts`, `service-utilities-types.d.ts` and `service-utilities.ts` are
functional, but not documented fully because this will go in a separate PR.

Connects-to: #70

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #70


hedss:
Connects-to: #85

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,871 @@
...
+/** The port that Github attempts to send events on. */
+const NotifyBotPort = 8399; // Not a listed registered port
+
+/**
+ * The RegExp to use to determine if a PR/Issue is connected to another. See the following links:
+ *   * https://help.github.com/articles/closing-issues-via-commit-messages/
+ *   * https://help.waffle.io/faq/waffle-workflow/use-waffles-connect-keyword-to-connect-prs-to-issues
+ */
+const IssueRefRE = new RegExp('(?:close[sd]?:?|fix(?:e[sd]{1})?:?|resolve[sd]?:?|' +
+	'connect(?:s|ed)?[\\s]+(?:to)?|connects-to?:?)[\\s]+(#[0-9]+)', 'i');
```
CameronDiver: Woah this is a big one! Can we get a couple examples of what it's supposed to match?

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,871 @@
...
+
+		// Create a new listener for Github with the right Integration ID.
+		const ghListener = this.addServiceListener('github', {
+			client: config.botName,
+			authentication: {
+				appId: config.githubApp,
+				pem: config.githubPEM,
+				type: GithubLogin.App
+			},
+			path: '/notifyhooks',
```
CameronDiver: I'd maybe move this out to a const.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,871 @@
...
+					this.logger.log(LogLevel.INFO, `Component version static, nothing to do for ${repo}`);
+				}
+			});
+
+			// We now have a set of components. For each of them, we now go through the process
+			// of following the Changelogs to find fixed issues.
+			// Any thrown error here needs to abort the run, as it means fundamentally something is
+			// wrong with the released versions.
+			return Promise.map(componentVersions, (component) => this.getNewPRs(component));
+		}).filter((component) => _.has(component, 'versions'))
```
CameronDiver: I'd newline this, I almost missed it as part of the above code block.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,871 @@
...
+		.then((components: ComponentTracker[]) => {
+			// We've a list of components and the PRs for each version.
+			// So, for each version we now harvest the relevant PRs and start the check
+			// backwards for Front URLs.
+			return Promise.map(_.filter(components, (comp) => comp.versions), (component: ComponentTracker) => {
+				return Promise.all(_.map(component.versions, (prs, key) => {
```
CameronDiver: Can `Promise.map` be used here?

lib/framework/procbot.ts
```diff
@@ -45,34 +44,31 @@ interface ServiceMap<T> {
  * * Retrieve configuration file from either the local FS or a ServiceEmitter
  */
 export class ProcBot {
+	/**
+	 * Retrieve the binary path for the Node dependencies.
+	 * @returns Promise containing a string denoting the absolute path.
+	 */
+	public getNodeBinPath = _.memoize((): Promise<string> => {
```
CameronDiver: I see what you're trying to do here, but [the lodash docs](https://lodash.com/docs/4.17.4#memoize) state that it will use the parameter as a key into an array. I'm not sure what will happen here, so I'd be wary at least.

lib/apis/githubapi-types.d.ts
```diff
@@ -40,6 +40,31 @@ export interface StatusEvent {
...
+	added: string[];
+	removed: string[];
+	modified: string[];
+	timestamp: string;
+}
+
+export interface PushEvent {
+	type: 'push';
+	action: string;
+	before: string;
```
hedss: I've put links to the Github API docs in the relevant sections of this file. I don't really want to go and re-document what's documented by GH.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,871 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import * as Promise from 'bluebird';
+import { Front, TopicConversations } from 'front-sdk';
+import * as GithubApi from 'github';
+import * as yaml from 'js-yaml';
+import * as _ from 'lodash';
+import * as GithubApiTypes from '../apis/githubapi-types';
```
hedss: I guess we could. I'll raise an issue for it though globally for later, rather than go through them all for this PR.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,871 @@
...
+	/** Hash of the file to retrieve. */
+	hash: string;
+	/** Owner of the repository the file exists in. */
+	owner: string;
+	/** Name of the repository the file exists in. */
+	repo: string;
+}
+
+/** Error object for passing to notification method. */
+interface NotifyBotError {
```
hedss: It's internal only, but yeah, I've modified it into a Typed Error descendant.

lib/bots/notifybot.ts
hedss: Sure, it's basically the Github Issue link keywords. Will embed a comment and a URL.

lib/bots/notifybot.ts
hedss: Done.

lib/bots/notifybot.ts
hedss: Oh, actually, there are already comments/URLs that specify this!

lib/bots/notifybot.ts
hedss: Done.

lib/framework/procbot.ts
hedss: There's no `resolver` parameter (I'm only passing the function used), so it doesn't apply.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+			}
+
+			// If there is an old component, use it to cull anything before that version.
+			if (component.oldVersion) {
+				changeParts = changeParts[0].split(component.incrementing ? component.oldVersion :
+					component.newVersion);
+			}
+			// We are guaranteed a new version, as it's already been verified.
+			changelogString = `## ${component.incrementing ? component.newVersion : component.oldVersion}` +
+				changeParts[0].split(component.incrementing ? component.newVersion : (component.oldVersion || ''))[1];
```
pimterry: It's very hard to tell what this logic is doing, especially with all the interspersed ternary operators and splits. Could we make it clearer? More understandable as something like:

* Split the changelog on new lines
* Find the indexes of the lines matching the new and old versions
* Grab the lines between
* Filter for version headings

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+			// We need to search the first entry in the array and split on every version
+			// so we know which versions to report back.
+			const versionEntries = changelogString.split('## ');
+
+			// Try and find a PR for each part. If we do, we store every PR we find
+			// in a version tracker.
+			const versionTracker: VersionPRTracker = {};
+			_.each(versionEntries, (versionData) => {
+				let verMatches: RegExpExecArray | null;
+				const versionMatch = versionData.match(/v([0-9]+\.[0-9]+\.[0-9]+)/);
```
pimterry: This should match with a `^## v...`, so we're sure it's not a version number from a changelog message

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+					}
+				}
+			});
+			component.versions = versionTracker;
+
+			return component;
+		});
+	}
+
+	// Retrieve a file from a repository given a full tree path and a commit hash.
```
pimterry: This comment is now duplicated

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+			return this.dispatchToEmitter(this.frontEmitterName, {
+				data: {
+					topic_id: shortenedTopic
+				},
+				method: this.frontApi.topic.listConversations
+			}).then((conversations: TopicConversations) => {
+				const convoList: string[] = [];
+				_.each(conversations._results, (conversation) => {
+					convoList.push(conversation.id);
+				});
```
pimterry: Easier as a map: `const convoList = conversations._results.map((conversation) => conversation.id)` or even `const convoList = _.map(conversations._result, 'id')`

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+			// Any thrown error here needs to abort the run, as it means fundamentally something is
+			// wrong with the released versions.
+			return Promise.map(componentVersions, (component) => this.getNewPRs(component));
+		})
+		.filter((component) => _.has(component, 'versions'))
+		.then((components: ComponentTracker[]) => {
+			// We've a list of components and the PRs for each version.
+			// So, for each version we now harvest the relevant PRs and start the check
+			// backwards for Front URLs.
+			return Promise.map(_.filter(components, (comp) => comp.versions), (component: ComponentTracker) => {
```
pimterry: Do you still need to filter here, when you already filter the components just above?

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+								`the ${prDetails.environment} has resolved the following issues that are attached ` +
+								'to this conversation:';
+						} else {
+							bodyMessage = `REGRESSED: Regression to version ${prDetails.deployedVersion} of ${prDetails.repo} ` +
+								`on the ${prDetails.environment} means the following issues attached to this ` +
+								'conversation are present again:';
+						}
+						_.each(topicIssues.relatedIssues, (issueURL) => {
+							bodyMessage += `\n${issueURL}`;
+						});
```
pimterry: `bodyMessage += '\n' + topicIssues.relatedIssues.join('\n')`

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+				// post on the PR that a version has been deployed, even though there may
+				// be no Front topics.
+				if (err.message !== 'Not Found') {
+					throw err;
+				}
+
+				return Promise.resolve([]);
+			});
+		};
+		const matchfrontTopics = (issue: GithubApiTypes.Issue, comments: GithubApiTypes.IssueComment[]) => {
```
pimterry: Front should probably be capitalized

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+const NotifyBotPort = 8399; // Not a listed registered port
+
+/**
+ * The RegExp to use to determine if a PR/Issue is connected to another. See the following links:
+ *   * https://help.github.com/articles/closing-issues-via-commit-messages/
+ *   * https://help.waffle.io/faq/waffle-workflow/use-waffles-connect-keyword-to-connect-prs-to-issues
+ * Additionally, this searches for the 'Connects-To:` keyword, which is specified by the Commit/PR guidelines.
+ */
+const IssueRefRE = new RegExp('(?:close[sd]?:?|fix(?:e[sd]{1})?:?|resolve[sd]?:?|' +
+	'connect(?:s|ed)?[\\s]+(?:to)?|connects-to?:?)[\\s]+(#[0-9]+)', 'i');
```
pimterry: This only matches issue links within the same repo, not cross-repo. Is that intentional? I don't think it's common, but there's definitely a few cases where a fix in a small module will be linked to an issue in the larger component repo.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+						topicIssues = results;
+						return this.retrieveTopics(topicIssues.topics);
+					}).then((conversations: string[]) => {
+						// Now we have all Front threads. Examine each of them for their IDs, and then post a private
+						// comment into them to note updated references.
+						let bodyMessage: string;
+						if (prDetails.incrementing) {
+							bodyMessage = `RESOLVED: Deployed version ${prDetails.deployedVersion} of ${prDetails.repo} on ` +
+								`the ${prDetails.environment} has resolved the following issues that are attached ` +
+								'to this conversation:';
```
pimterry: Here we assume that every issue that's linked to a PR is completely resolved by a PR, right? That means if I create any PR with `connects to: #largerissue`, when it's deployed it'll immediately put 'RESOLVED' on the issue and every related Front conversation. What about all the issues that aren't resolved by a single PR? That's got to be a substantial percentage, and probably a majority of non-trivial issues.

This would be largely mitigated by only matching `fixes`/`closes` - I think it's `connects to` that's particularly challenging here.

hedss:
@pimterry Sorry, yes, it was very large. I've taken your comments onboard.

lib/framework/procbot.ts
hedss: True, removed.

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
...
+ * - A standard way of getting contextualised workers.
+ *
+ * In exchange you agree to:
+ * - provide a `emitData` method which takes a ServiceEmitContext and returns a promise.
+ * - provide a `verify` function which should check that incoming data isn't spoofed.
+ * - provide a `serviceName` getter, which because it is based on file path cannot be inherited away.
+ * - provide a `apiHandle` getter, which should return the underlying object that executes the requests.
+ * - enqueue your events with `context` and `event`, as per `UtilityServiceEvent`.
+ */
+export abstract class ServiceUtilities<T> extends WorkerClient<T> implements ServiceListener, ServiceEmitter {
```
hedss: Yeah, agree. This is part of the work I've had to pull in from https://github.com/resin-io-modules/resin-procbots/pull/245 after previous changes had been made that stopped many of the services working as anything other than a message bus for SyncBot. Whilst I don't disagree, I'd like to address these in that PR rather than this.

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
...
+
+	/**
+	 * Emit data to the service.
+	 * @param data  Service Emit Request to send, if relevant.
+	 * @returns     Details of the successful transmission from the service.
+	 */
+	public sendData(data: ServiceEmitRequest): Promise<ServiceEmitResponse> {
+		const context = data.contexts[this.serviceName] as ServiceEmitContext;
+		if (!context) {
+			return Promise.resolve({
```
hedss: This is down to Client Bots being able to dispatch emit commands to multiple services at the same time using `dispatchToAllEmitters`. This means we can potentially end up in a situation where some dispatches have been successful, and some have failed. It's then up to the client to determine if any of the failures are actually a problem.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+/** The port that Github attempts to send events on. */
+const NotifyBotPort = 8399; // Not a listed registered port
+
+/**
+ * The RegExp to use to determine if a PR/Issue is connected to another. See the following links:
+ *   * https://help.github.com/articles/closing-issues-via-commit-messages/
+ *   * https://help.waffle.io/faq/waffle-workflow/use-waffles-connect-keyword-to-connect-prs-to-issues
+ * Additionally, this searches for the 'Connects-To:` keyword, which is specified by the Commit/PR guidelines.
+ */
+const IssueRefRE = new RegExp('(?:close[sd]?:?|fix(?:e[sd]{1})?:?|resolve[sd]?:?|' +
```
hedss: That's fabulous and bookmarked. Thanks!

lib/bots/notifybot.ts
hedss: Done.

hedss: Fixed.

hedss: Fixed.

hedss: Yes, very intentional. The NB spec. says it will only match against local issues and those from HQ with the `hq:` tag.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+	'connect(?:s|ed)?[\\s]+(?:to)?|connects-to?:?)[\\s]+(#[0-9]+)', 'i');
+/** The RegExp used to determine if an HQ issue is linked to from a component issue. */
+const HqRE = /hq:[\s]+https:\/\/github.com\/resin-io\/hq\/issues\/([0-9]+)/i;
```
hedss: That's true. I'd anticipated that only handling the first link in both cases. But actually, this isn't a good fit. Will change.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+	/** Github ServiceListener name. */
+	private githubListenerName: string;
+	/** Github ServiceEmitter name. */
+	private githubEmitterName: string;
+	/** Front ServiceEmitter name. */
+	private frontEmitterName: string;
+	/** Front user to use when commenting on Front conversations. */
+	private frontUser: string;
+	/** Front API key for access to a specific Front service. */
+	private frontApiKey: string;
```
hedss: Good spot.

lib/bots/notifybot.ts
hedss: Nope, looks like I forgot to throw it out.

lib/bots/notifybot.ts
hedss: This changes how we'd specced the bot though. I think we need to discuss this again and decide how we're going to determine the actual fixed Issue. From our Commit/PR spec.: 'We don't use the github fixes, closes et al. keywords, because these result in the associated issue to be closed as soon as the PR is merged.' Issues aren't closed until the PR is deployed, but NB is only going to activate when a Keyframe has just been deployed to an environment, which is the first time an Issue can actually said to be closed.

So, we have a problem. We're going to need to determine the Issue that's actually closed. We could, for example, say that the body for a PR must *only* include Issues that are resolved by this PR. That way any further comments on the PR can include cross-referenced issues that aren't actually resolved. Subsequently, Issues for a PR that cross-reference an HQ Issue should only connect them in the body? (This is actually slightly mitigated because NB will only follow HQ issues referenced by `hq:`). Or we could use a separate keyword that won't close the Issue but isn't one we use for anything else.

cc @lekkas

lib/bots/notifybot.ts
hedss: Wood, trees. :)

lib/bots/notifybot.ts
hedss: Actually, having spoken to Tim again, I'll just change the wording. There's potential that an agent may have to trawl through lots of Issues commented on in Front, but if that becomes a problem we'll refine it.

lib/bots/notifybot.ts
hedss: Yup.

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+	 * @param component  Component information object.
+	 * @returns          An updated component information object.
+	 */
+	private getNewPRs(component: ComponentTracker): Promise<ComponentTracker> {
+		// Get the changelog for the component from its `master`, as this file
+		// *should* include all given versions. If it doesn't, then something is broken
+		// or the keyframe is incorrect.
+		const repo = this.getRepoDetails(component.repo);
+		const validVersion = (data: string, version: string) => {
+			return data.indexOf(version) !== -1;
```
hedss: Yeah, that's a better way of doing it.

lib/bots/notifybot.ts
hedss: Hmm, VC whinged about it, which has surprised me (as it can be `string | null`).

lib/bots/notifybot.ts
```diff
@@ -0,0 +1,918 @@
...
+				// A new version less than an older version is a version down.
+				component.incrementing = false;
+			} else if (component.newVersion > component.oldVersion) {
+				component.incrementing = true;
+			}
+
+			// If there is an old component, use it to cull anything before that version.
+			if (component.oldVersion) {
+				changeParts = changeParts[0].split(component.incrementing ? component.oldVersion :
+					component.newVersion);
```
hedss: Yeah, I'd agree with that. :/

lib/services/service-utilities.ts
pimterry: Personally, I'd rather fail by default, and find a nice way to let clients ignore certain errors/emitters if they need to, but that doesn't really matter for this PR - sounds like this is fine here for now.

lib/bots/notifybot.ts
hedss: We had a discussion about this in the process meeting yesterday. We're going to go ahead and (at some point) go for a module that parses changelogs based.

pimterry: > We're going to go ahead and (at some point) go for a module that parses changelogs based.

...based on what? 

Interesting thought that occurs to me: why bother parsing the changelog itself? We can get close with regexes here, but it's not perfect - it's a plain text human readable document with all the fuzzy mess that entails. How hard would it be get the commits + messages directly, and just pull this data from them? There you've got much cleaner & more structured data. I can imagine in some places that might be tricky to quickly grab though...

Could easily start getting versionbot to add hidden metadata in changelogs to make this easier instead? You should be able to put in HTML comments with arbitrary data inline, that don't actually appear in the human readable doc, and are _very_ unlikely to conflict with : `<!--VB { version: '1.2.3', pr: '#123', user: 'pimterry' } -->`



Alternatively, there's a way more long winded approach where we treat human readable changelogs as a final artefact, but generate a machine readable one first, so we have both available.

lib/bots/notifybot.ts
hedss: Because of the split the start of the line should be `v1.2.3 - Blah blah blah`, but it does need a `^` prefixed.

lib/bots/notifybot.ts
hedss: Based on a machine-readable version, no idea what happened to the rest of my sentence! So yes, a similar idea. Initially, the reason for processing the changelogs here instead of the commits was to only include things that were in 'release note' format, rather than commits that didn't have an entry in the changelog (although for most of our PRs, that doesn't occur).

lib/bots/notifybot.ts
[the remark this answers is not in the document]
pimterry: Hmm, you didn't actually do this - it looks like you've dropped the `:?` entirely, except for `connects-to:`, so the current regex no longer allows `fixes: #123` for example. Is that intentional?

lib/bots/notifybot.ts
```diff
@@ -762,35 +767,44 @@ export class NotifyBot extends ProcBot {
...
+			// contains all entries between the old and new version.
 			changelogString = `## ${component.incrementing ? component.newVersion : component.oldVersion}` +
 				changeParts[0].split(component.incrementing ? component.newVersion : (component.oldVersion || ''))[1];
 
-			// We need to search the first entry in the array and split on every version
-			// so we know which versions to report back.
+			// Split the changelog based on version entries. This ensures we end up with
+			// an array, each entry being a single version set of changes, for example:
+			//  "v.1.2.3 - 2018-08-02\n\n* This does something else #124 [Author]"
+			//  "v.1.2.2 - 2018-08-01\n\n* This does something #123 [Author]"
```
pimterry: There's an extra `.` in these version numbers, between the `v` and the `1`

hedss: Fixed.

lib/bots/notifybot.ts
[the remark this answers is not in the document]
hedss: Yep, stuffed it up. It now resolves to:
`((?:close[sd]?|fix(?:e[sd]{1})?|resolve[sd]?):?|connect(?:(?:s|ed)?\s+(?:to)?:?|s-to:))\s+(#[0-9]+)`
Which accepts optional trailing semi-colon, apart from on `connects-to: #` where it must exist (VB Waffleboard format).

PR #245, opened 2017-08-11
173 messenger refactor

sqweelygig:
***THIS PR SHOULD NOT BE MERGED***

I'm creating this PR to assist in providing structured feedback on the current state of SyncBot.

At the moment it creates new threads between flowdock and forums.  To add to this is front, new comments, backporting and rebasing (not necessarily in that order)


hedss:
@sqweelygig Sorry, I just have not had time to get round to this this week, been an utterly manic week. I'll be doing this first thing Monday though (obviously won't be a full review, just architecture).

package-lock.json
```diff
@@ -0,0 +1,5893 @@
+{
```
hedss: NPM 5's `package-lock.json` is currently (unless I've missed something) a bit 'up in the air' (https://github.com/npm/npm/issues/16866). I'd rather not commit the lock file until it's extremely clear as to what the flow is (there doesn't appear to be consistency).

Additionally, we may only want to eventually commit to the application version of ProcBots, and not the library. It needs discussion.

lib/services/errors/context-absent.ts
```diff
@@ -0,0 +1,2 @@
+import TypedError = require('typed-error');
```
hedss: I'm against this on the principal that it's a very specific error class, and we'd end up with them dotted all over the place. I'm far more in favour of including any error type that's associated with a class as a definition in the class itself (look at https://github.com/resin-io-modules/resin-procbots/blob/master/lib/services/github.ts#L60 as an example).
For that reason, I think there should be a `ProcBotError` type, that exists in `procbot.ts`. This would include both `code` and `message`, where code is from a const enumerated type defined in `procbot-types.ds`. This is therefore one such `ProcBotError` error type.

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
...
+	/**
+	 * Emit data to the service.
+	 * @param data  Service Emit Request to send, if relevant.
+	 * @returns     Details of the successful transmission from the service.
+	 */
+	public sendData(data: ServiceEmitRequest): Promise<ServiceEmitResponse> {
+		const context = data.contexts[this.serviceName] as ServiceEmitContext;
+		if (!context) {
+			return Promise.resolve({
+				err: new ContextAbsentError(`No ${this.serviceName} context`),
```
hedss: As previous comment, this could by a specific `ProcBotError`, eg. `new ProcBotError({ code: 123, message: `No ${this.serviceName} context`)})`.

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
...
+ * - A standard way of getting contextualised workers.
+ *
+ * In exchange you agree to:
+ * - provide a `emitData` method which takes a ServiceEmitContext and returns a promise.
+ * - provide a `verify` function which should check that incoming data isn't spoofed.
+ * - provide a `serviceName` getter, which because it is based on file path cannot be inherited away.
+ * - provide a `apiHandle` getter, which should return the underlying object that executes the requests.
+ * - enqueue your events with `context` and `event`, as per `UtilityServiceEvent`.
+ */
+export abstract class ServiceUtilities<T> extends WorkerClient<T> implements ServiceListener, ServiceEmitter {
```
hedss: @pimterry pointed out that this isn't really utilities in the strict sense of the word, as Services inherit from it if they want to use it. I don't really like `AbstractService` as a Service doesn't *have* to inherit from it. `ServiceParent`? `ServiceFramework`? I have no idea! 😺

lib/services/service-utilities-types.d.ts
```diff
@@ -0,0 +1,29 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import { WorkerEvent } from '../framework/worker';
+import { ServiceEvent } from './service-types';
+
+/** Data object the ensures the existence of context and event for serviceUtilities. */
+export interface UtilityServiceEvent extends ServiceEvent {
```
hedss: The event should probably use the same nomenclature as the class that uses it, `ServiceUtilityEvent`, maybe.

lib/services/service-utilities-types.d.ts
```diff
@@ -0,0 +1,29 @@
...
+	context: any;
+	type: string;
+}
+
+/** Incoming event object that guarantees some details about the data for serviceUtilities generic handling. */
+export interface UtilityWorkerEvent extends WorkerEvent {
```
hedss: `ServiceUtilityWorkerEvent`?

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
...
+	/** A singleton express instance for all web-hook based services to share. */
+	private static _expressApp: express.Express;
```
hedss: We need to stop using `_` for private variable naming (https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines), I *think* I've managed to get rid of all instances in my code apart from `_botname` and `_serviceName`, but they need to go, too. It's way too easy to also confuse this for a `lodash` method. :smiley:

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
...
+		const created = new Worker<T>(context, this.removeWorker);
+		this.workers.set(context, created);
+		return created;
+	}
+
+	/**
+	 * Create or retrieve the express app.
+	 * @returns  Express server app.
+	 */
+	protected get expressApp(): express.Express {
```
hedss: I think there needs to be the ability to pass an already existing Express instance in here, similar in style to what the `front-sdk` package does.

lib/services/service-utilities.ts
```diff
@@ -0,0 +1,204 @@
+/*
```
hedss: In general, I love this!

typings/flowdock.d.ts
```diff
@@ -10,5 +10,9 @@ declare module 'flowdock' {
 		public stream(ids: string[]): FlowdockStream;
 		public on(event: string, callback: (error: Error) => void): void;
 		public removeListener(event: string, callback: (error: Error) => void): void;
+		public _request(
```
hedss: No underscore.

lib/services/front.ts
```diff
@@ -1,357 +1,127 @@
...
+// export class FrontService extends ServiceUtilities<string> implements ServiceListener, ServiceEmitter {
+// 	private static _serviceName = path.basename(__filename.split('.')[0]);
+//
+// 	/** Underlying SDK object that we route requests to */
+// 	private session: Front;
+//
+// 	constructor(data: FrontConnectionDetails, listen: boolean) {
+// 		super();
+// 		this.session = new Front(data.token);
+// 		if (listen) {
```
hedss: I don't think there's much need for a separate `startListening()` method. The constructor for most services is going to be small enough that adding the code under `if (listen) {` is fine.

lib/services/front.ts
```diff
@@ -1,357 +1,127 @@
...
+// import { ServiceEmitter, ServiceListener } from './service-types';
+// import { ServiceUtilities } from './service-utilities';
+//
+// export class FrontService extends ServiceUtilities<string> implements ServiceListener, ServiceEmitter {
+// 	private static _serviceName = path.basename(__filename.split('.')[0]);
+//
+// 	/** Underlying SDK object that we route requests to */
+// 	private session: Front;
+//
+// 	constructor(data: FrontConnectionDetails, listen: boolean) {
```
hedss: I'd argue that `listen` could just be another option in `FrontConnectionDetails`. Why have separate constructor parameters?

lib/services/discourse.ts
```diff
@@ -17,211 +17,80 @@
...
-					url: getTopic.uri,
-					user: details.post.username,
-				},
-				text: metadata.content,
-				title: details.topic.title,
-			};
-		});
+	private connectionDetails: DiscourseConnectionDetails;
+
+	constructor(data: DiscourseConnectionDetails, listen: boolean) {
```
hedss: As for comments on Front, just have `listen` as an option in `DiscourseConnectionDetails`.

lib/services/discourse.ts
```diff
@@ -17,211 +17,80 @@
...
+		super();
+		// #203: Verify connection data
+		this.connectionDetails = data;
+		if (listen) {
+			this.startListening();
```
hedss: As for comments in Front, remove `startListening()` method and just embed the code here.

lib/services/discourse.ts
```diff
@@ -17,211 +17,80 @@
...
-					},
-					payload: {
-						category: data.toIds.flow,
-						raw: `${data.text}\n\n---\n${Messenger.stringifyMetadata(data)}`,
-						title,
-						unlist_topic: data.hidden ? 'true' : 'false',
-					}
+	public request(requestOptions: UrlOptions & RequestPromiseOptions): Promise<DiscourseResponse> {
+		// This type massages request-promise into bluebird
+		return new Promise((resolve) => {
```
hedss: If we're using `request-promise`, why is there a `new Promise` here instead of just returning the result of `request`?

lib/services/flowdock.ts
```diff
@@ -16,295 +16,83 @@ limitations under the License.
...
+export class FlowdockService extends ServiceUtilities<string> implements ServiceEmitter, ServiceListener {
 	private static _serviceName = path.basename(__filename.split('.')[0]);
 	private session: Session;
-	private data: FlowdockConstructor;
+	private org: string;
 
-	public constructor(data: FlowdockConstructor, listen = true ) {
-		super(listen);
-		this.data = data;
+	constructor(data: FlowdockConnectionDetails, listen: boolean) {
```
hedss: As per Front comments, `listen` could be an option in `FlowdockConnectionDetails`.

lib/services/flowdock.ts
```diff
@@ -16,295 +16,83 @@ limitations under the License.
...
+		super();
 		this.session = new Session(data.token);
+		this.org = data.organization;
+		if (listen) {
+			this.startListening();
```
hedss: Drop `startListening()` method and add the code into this block.

lib/services/front.ts
```diff
@@ -1,357 +1,127 @@
...
-}
-
-/**
- * Build this class, typed as a message service.
- * @returns  Message Service object, ready to convert events.
- */
-export function createMessageService(data: FrontConstructor): Messenger {
-	return new FrontService(data, false);
-}
+// /*
```
hedss: I've got a slightly modified (maybe) version of this in the NotifyBot PR: https://ghttps://github.com/resin-io-modules/resin-procbots/pull/243ithub.com/resin-io-modules/resin-procbots/pull/243

lib/services/messenger.ts
```diff
@@ -15,300 +15,118 @@
...
-		if (listener) {
-			// Allow the sub-constructor to set up sessions, etc, before listening
-			process.nextTick(this.listen);
+		this.translators = {};
+		_.forEach(data.subServices, (subConnectionDetails, serviceName) => {
+			this.logger.log(LogLevel.INFO, `---> Constructing '${serviceName}' translator.`);
+			this.translators[serviceName] = Translator.createTranslator(serviceName, subConnectionDetails, data.dataHubs);
+		});
+		if (listen) {
+			this.startListening(data.subServices);
```
hedss: Move the `startListening()` method into this body.

package-lock.json
sqweelygig: That one slipped through, not deliberate, thanks for catching.  For now I have put it in the .gitignore.

sqweelygig: Note to self: done.

lib/services/errors/context-absent.ts
sqweelygig: Note to self: Done.

lib/services/service-utilities.ts
sqweelygig: Note to self: dopamine.

lib/services/service-utilities.ts
sqweelygig: Note to self: Done, ServiceScaffold.

lib/services/service-utilities.ts
sqweelygig: Note to self: Done.

lib/services/service-utilities.ts
sqweelygig: Note to self: Done.

lib/services/service-utilities.ts
sqweelygig: Note to self: Done

lib/services/service-utilities-types.d.ts
sqweelygig: Note to self: Done

lib/services/service-utilities-types.d.ts
sqweelygig: Note to self: Done.

typings/flowdock.d.ts
sqweelygig: Well, this is actually just a typing around someone else's SDK, which reveals a slightly different error of using someone else's 'private by convention' methods, which I'm fixing.

sqweelygig: Note to self: Done

lib/services/front.ts
sqweelygig: The Front service has currently been put temporarily out of scope, as indicated by the massive commenting, and will be coming back into the fold in a couple of days.

lib/services/front.ts
sqweelygig: The Front service has currently been put temporarily out of scope, as indicated by the massive commenting, and will be coming back into the fold in a couple of days.

lib/services/front.ts
sqweelygig: The Front service has currently been put temporarily out of scope, as indicated by the massive commenting, and will be coming back into the fold in a couple of days.

lib/services/discourse.ts
sqweelygig: To discuss.  The reason for it being like this is that this information can be derived from whether you want a ServiceListener or a ServiceEmitter and is therefore available to the typed factory methods.  By allowing the property to be set at the higher level, in the case of DiscourseConnectionDetails at env var level, it leads to the possibility of a listener that doesn't listen and more generally passes up the stack details that aren't required.

lib/services/discourse.ts
sqweelygig: Note to self: Done

lib/services/discourse.ts
sqweelygig: Note to self: Done

lib/services/flowdock.ts
sqweelygig: To discuss.

lib/services/flowdock.ts
sqweelygig: Note to self: Done

lib/services/messenger.ts
sqweelygig: Note to self: Done

lib/services/discourse.ts
sqweelygig: Or, it would seem, to sleep on it.  I woke up this morning know how to approach this.  The constructor objects now have an optional boolean `deaf`, which if truthy skips the listen code.

PR #268, opened 2017-09-21
Handle spawned process errors as errors, which allows us to report versionist errors more nicely.

Page-:
Change-type: patch


hedss:
Sod, I hit 'Update branch' totally by mistake. Sorry. :/

PR #281, opened 2017-10-05
SyncBot: Make metadata encoding invisible

sqweelygig:
Pinging Sonya and Gergely, one of them can confirm that the bullets below describe a desired feature.
Pinging Cam, Pim and Hedss to confirm that the code implements these bullet points and is well styled.

* Use empty links, either markdown or html, to encode metadata instead of footers

Change-Type: patch


sonyagreen:
I'm all for no more footers. I can't confirm empty links to encode metadata is the way to do it.

lib/services/messenger/translators/translator-scaffold.ts
```diff
@@ -27,6 +27,10 @@ import {
 	ResponseConverters, TranslatorErrorCode, TranslatorMetadata,
 } from './translator-types';
 
+export enum MetadataEncoding {
+	hiddenHTML, hiddenMD, char,
```
hedss: Can we have these starting with a Capital? It's easier to differentiate them from locally scoped vars, then.
Also, I'd strongly consider renaming `char`, which whilst not a reserved keyword in TS, is for future use in JS.

sqweelygig:
Added Alex to the list of reviewers to confirm the desirability of invisible metadata over footers.

sqweelygig:
* This PR will cause metadata to be encoded invisibly.

@hedss @CameronDiver or @pimterry: Approve if this code does ^^that bullet point^^ and is well styled.
I'm prepared to take the risk and not wait for explicit confirmation of it being a good idea, because there were Flowdock threads that I can't find right now, but if @alexandrosm or @imrehg were to chip in to say "Invisible metadata is much better than camouflaged, that's great!" I'd appreciate it.

sqweelygig:
Adding @LucianBuzzo to this PR since he's been working in this repo recently.

PR #282, opened 2017-10-05
SyncBot: Report possible solutions alongside errors.

sqweelygig:
Going to ping Sonya in to confirm that the bullets below form a desirable solution, even though she didn't request it, she's possibly the best placed.
Pinging Pim, Cam and Hedss to confirm that this implements the bullets below and is well styled.

* Allow the environment to be configured to provide solutions to errors.

Change-Type: patch


sonyagreen:
>Allow the environment to be configured to provide solutions to errors.

I don't know what that means. Environment where? Solutions  to what errors?

sqweelygig:
Okay, cool.  So with this code the server can be configured to translate a service's error messages into more user friendly words.  So, for example when Discourse says "403 - you are not permitted to view the requested resource.", SyncBot will, instead of reporting the error verbatim, say "Discourse reports permissions error.  Try checking your usernames match and that your password has been provided"

hedss:
@sqweelygig Why wouldn't you just put errors in the source? Are these ever likely to change? I'd point to the current CloudFormation envvar issues as reasons why having copious amount of envvars is a bad idea.

sqweelygig:
The problem with the service, or translator, knowing about it's own errors is that they do not know how things are configured.  No where in the source code that handles Discourse does it know that we should be inputting our tokens in the 1-1 history on Flowdock.  The only place that knows how this is setup is the setup.

hedss:
So explicitly check for the different service errors in SyncBot, then? I don't see why this couldn't be an object defined at the top of the source for it.

sqweelygig:
We provide our Discourse credentials using the Flowdock 1-1 history.  It is not the SyncBot codebase that knows this, but the SYNCBOT_DATAHUB_CONSTRUCTORS environment variable.  It is therefore not the SyncBot codebase that understands how to resolve problems with Discourse credentials.
@hedss I think we need to go synchronous on this, with my apologies to your processor.

sqweelygig:
I've just put this in to Architecture.  I'm not really happy with massive-chunk-of-JSON, nor external-file, which are the obvious options under the current keyframe scripts.  I've added to the mix keyframe-to-understand-YML.  We'll see what happens.

sqweelygig:
@hedss this is now good to reconsider.  I'm content enough to use massive JSON dump for now, and look forwards to YAML encoding.

lib/bots/syncbot.ts
```diff
@@ -94,9 +110,9 @@ export class SyncBot extends ProcBot {
 					}
 					return emitResponse;
 				})
-				// Then report that we have passed the message on
+				// Once we reach the end of our actions we don't actually care about the response and should void it away
 				.then(() => {
```
hedss: Should get rid of the `then` if it has no body, and shouldn't be returning the `emitResponse` from the previous promise. Currently it's confusing as it looks like something was going to be done with the body, but was forgotten.

lib/bots/syncbot.ts
```diff
@@ -106,13 +106,10 @@ export class SyncBot extends ProcBot {
 							service: to.service,
 							flow: to.flow,
 							thread: threadId,
-						}, messenger, data);
+						}, messenger, data)
+						// The upstream expectations are for a promise that resolves to void when complete.
+						.thenReturn();
```
hedss: Whilst slightly pedantic, we use `.return` in every other CS and TS code, so I'd rather we use it here, too. Ta!

PR #283, opened 2017-10-05
SyncBot WIP: Capture message details for future work

sqweelygig:
This code should have no UX implications, but allows for capture of webhook details without a development instance of intercom.
Pinging Hedss, Cam and Pim to confirm that this does what I say and is well styled.

* Add option to capture details of the message form Front, to inspect intercom messages

Change-Type: patch


hedss:
Looking at the API docs for Intercom, Access Tokens can be created/revoked easily for individuals? Is this difficult to request, work with?
Won't this change spam with absolutely everything from Front, not just Intercom messages?

sqweelygig:
The problem with that is that I want to see how Front emits the intercom message, I'm hoping that the Front service won't have to deal with the Intercom API to get decent user handles.

lib/services/messenger/translators/front.ts
```diff
@@ -45,6 +45,18 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
 	 * @returns        username or handle from Front.
 	 */
 	private static extractAuthorDetails(message: Message): { username: string, internal: boolean } {
+		// TODO: Remove this code, once I've captured an example message from intercom.
+		// This process should be much easier than setting up a development instance of intercom
+		// And the findings should be documented somewhere
+		if (process.env.SYNCBOT_INTERCOM_INVESTIGATION) {
+			console.log('---Message object begins---');
```
hedss: This should use the logger rather than `console.log` everywhere.

hedss:
I'm not sure why this still isn't achievable using Front sandbox and an Intercom Access Token, but it shouldn't affect the rest of the code, I guess.

sqweelygig:
Ah-hah!  Right, that just clicked.  I will try that approach, leave this PR alone until then please.  I'm going to remove all reviewers to indicate that.

Issue #284, opened 2017-10-06
Circle doesn't tag new versions when merged to master


hedss:
And I'm fed up of doing it manually.

PR #285, opened 2017-10-06
Ensure that when merged to master, images are tagged and pushed with new version to Docker hub

hedss:
Connects-to: #284
Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #284


circle.yml
```diff
@@ -34,3 +34,11 @@ deployment:
           docker push ${DOCKER_IMAGE}:${CIRCLE_TAG}
         fi
 
+        # On master merges, ensure we tag the new version.
+        [[ "${CIRCLE_BRANCH}" == "master" ]] || exit 0
```
brownjohnf: Is there a reason you don't want to run this on other branches? It'll generate a nice tag like v2.3.0-<sha>, which allows one to pull/run builds from branches.

hedss: That's a good point. I'll remove the `master` only check.

hedss:
@lekkas Yep, `git` is actually already included, so didn't need the line anyway!

brownjohnf:
I think the reason we have to add it to some of our jobs is that we don't have it when we use a custom build env image on circle, like when we need docker-in-docker

PR #289, opened 2017-10-06
Run build step in Docker build process

### LINKED ISSUE #288: Build source in Dockerfile
At the moment, you need to run `gulp build` and commit the built changes. If we're using a Dockerfile, then the build step can just happen in the Docker build and we can remove the build files from the repo.

LucianBuzzo:
Fixes #288

change-type: patch


.gitignore
```diff
@@ -44,3 +44,6 @@ _dev_notes
 
 # package-lock
 package-lock.json
+
+# build files
+build
```
hedss: Newline! :)

LucianBuzzo:
@hedss 
Only the files necessary to run Procbots end up in the final image.
I've kept `package.json` as it is useful for checking versions when one is debugging.
The working directory looks like this
```sh
drwxr-xr-x   8 root root  4096 Oct 13 11:38 .
drwxr-xr-x   8 root root  4096 Oct 13 11:22 ..
drwxr-xr-x   2 root root  4096 Oct 13 11:38 bin
drwxr-xr-x   6 root root  4096 Oct 13 11:38 build
drwxr-xr-x 628 root root 20480 Oct 13 11:39 node_modules
-rw-rw-r--   1 root root  2220 Oct 13 11:38 package.json
```

PR #291, opened 2017-10-06
Ensure that the connection keywords in a Github Issue are correctly handled

hedss:
Change-type: patch


hedss:
This is the last issue before rolling out NB (I hope!).

lib/bots/notifybot.ts
```diff
@@ -182,8 +182,9 @@ const NotifyBotPort = 8399; // Not a listed registered port
  *   * https://help.waffle.io/faq/waffle-workflow/use-waffles-connect-keyword-to-connect-prs-to-issues
  * Additionally, this searches for the 'Connects-To:` keyword, which is specified by the Commit/PR guidelines.
  */
-const IssueRefRE = new RegExp('(?:close[sd]?|fix(?:e[sd]{1})?|resolve[sd]?):?|' +
-	'connect(?:(?:s|ed)?\\s+(?:to)?:?|s-to:)\\s+(#[0-9]+)', 'gi');
+
+const IssueRefRE = new RegExp('(?:(?:close[sd]?|fix(?:e[sd]{1})?|resolve[sd]?|connect(?:(?:s|ed)?\\s+(?:to)?)):?|' +
+	'connects-to:)\\s+(#[0-9]+)', 'gi');
```
CameronDiver: What about when the connects to link is an issue in a separate repository? Or is it assumed that a mirrored issue will be opened in the repo in which the PR is opened?

hedss: Connections are only allowed in the same repo or as an HQ link, from the NB spec.

PR #305, opened 2017-10-18
SyncBot: Use aliases when posting to Flowdock

sqweelygig:
This is to revert a setup requirement that I'd accidentally introduced in the refactor.

* No longer require every user to provide their Flowdock token.
* Use a generic token and aliases when posting to Flowdock.

Change-Type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


sqweelygig:
Broadened the net a little on potential reviewers for this one.  What this change does is mean that when SyncBot goes looking for details to make a message suitable for Flowdock it goes looking in the constructor object, taken from environment, rather than (via DataHubs, see below) to a person's flowdock history.
DataHubs are a way of configuring into the environment a list of places to seek information, such as tokens.  They are used for situations, such as a resineer's discourse token, where we know where to look as we're configuring the instance, but not what the values will be.

Issue #204, opened 2017-07-17
SyncBot: Front to verify received events


sqweelygig:
At the moment this performed by only using the event id from the webhook and GETting everything else, but shared secrets would improve this.

hedss:
Is the Front Service not using the Front SDK for this? If not, it probably should be. It verifies all events automatically.

hedss:
This is part of #200

PR #309, opened 2017-10-24
SyncBot, ProcBots: Use Front SDK to verify and expand events.

sqweelygig:
* Front service now uses the registration of interest method in the Front SDK.
    * This automatically verifies against a shared secret
    * This automatically expands the event preview.

Change-Type: patch
Connects-To: #204 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #204


lib/services/front.ts
```diff
@@ -37,25 +37,31 @@ export class FrontService extends ServiceScaffold<string> implements ServiceList
...
-					cookedEvent: {},
-					rawEvent: formData.body,
-					source: FrontService._serviceName,
-					type: formData.body.type,
-				});
-				response.sendStatus(200);
+			this.session.registerEvents({
+				hookPath: `/${listenerData.path || FrontService._serviceName}`,
+				server: this.expressApp,
+			}, (error: Error | null, event?: Event) => {
```
LucianBuzzo: I know this is pre-existing, but why does this use a callback instead of a promise? Looking at the internal code it looks like bothe `registerEvents` and `httpCall` both use promises internally but use callbacks in their call signature.

lib/services/front.ts
```diff
@@ -37,25 +37,31 @@ export class FrontService extends ServiceScaffold<string> implements ServiceList
...
+				if (!error && event) {
+					this.queueData({
+						context: event.conversation.id,
+						cookedEvent: {},
+						rawEvent: event,
+						source: FrontService._serviceName,
+						type: event.type,
+					});
+				}
```
LucianBuzzo: There should be some kind of error handling here, even if it's just `console.error`

lib/services/front.ts
```diff
@@ -70,10 +76,10 @@ export class FrontService extends ServiceScaffold<string> implements ServiceList
 	}
 
 	/**
-		* Verify the event before enqueueing.  For now uses the naive approach of returning true.
+		* Verify the event before enqueueing. Trusts the SDK to perform verification.
```
LucianBuzzo: NIT: Weird tab indentation here

lib/services/front.ts
sqweelygig: This uses a callback because it is registering a handler that may occur many times in response to external events, rather than a Promise which should only be completed once.

lib/services/front.ts
sqweelygig: Done

sqweelygig:
Just a quick note, mainly to self.  This currently does not work because of changes made to the shared express app initialisation because Discourse requires that the raw string be verified prior to JSON parsing.  Making a PR in the front-sdk module to support raw string verification.

Issue #217, opened 2017-07-21
SyncBot: Support having a Front conversation in multiple inboxes.


sqweelygig:
At the moment a Front conversation that ends up in more than one inbox makes SyncBot fail silently.  It should just work.

sqweelygig:
https://www.flowdock.com/app/rulemotion/r-process/threads/_OteIn8wFuNkgtxwkdxWZsJtPoo

PR #313, opened 2017-10-26
SyncBot: Front events now possible in multiple inboxes.

sqweelygig:
**PLEASE NOTE** I have not been able to replicate the bug in my development environment due to not having aliased email accounts.  Two factors offset this however.  Firstly steps have been taken to reduce the occurrences of this bug in our production setup.  Secondly these changes are reasonably provable.

* Translators now create a list of messages from one event.
* Front translator expands a multiple inbox event into multiple messages.

Change-Type: patch
Connects-To: #217

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #217


lib/services/messenger.ts
```diff
@@ -64,8 +64,10 @@ export class MessengerService extends ServiceScaffold<string> implements Service
 					listenerMethod: (_registration: ServiceRegistration, event: ServiceScaffoldEvent): Promise<void> => {
 						// This translates and enqueues any events that match a registered interest.
 						if (_.includes(this.eventsRegistered, subTranslator.eventIntoMessageType(event))) {
-							return subTranslator.eventIntoMessage(event)
-							.then(this.queueData);
+							return subTranslator.eventIntoMessages(event)
+							.then((messages) => {
+								_.forEach(messages, this.queueData);
```
brownjohnf: Does typescript automatically return the last line? Otherwise, this may not have completed before it's used.

hedss: This (correct me if I'm wrong Andrew) should end up queuing an event, which doesn't wait for a response but pushes the data into an async event queue. So, it doesn't actually need to wait for completion.

hedss: This (correct me if I'm wrong Andrew) should end up queuing an event, which doesn't wait for a response but pushes the data into an async event queue. So, it doesn't actually need to wait for completion.

sqweelygig: No correction needed, Hedss is entirely correct.

sqweelygig:
I have built and tested it for single-inbox messages, and will do a paranoia check this before merging.

Issue #297, opened 2017-10-09
SyncBot: Should handle squashing rooms


sqweelygig:
So the idea is that, for example, forums/troubleshooting and forums/resin.io both go into flowdock/s/public/forums and front/s/forums.  Right now that might be achievable by creating several equivalencies, but if anything attempted the reverse journey it would go ballistic.  One potential solution, which would also feed nicely into the hashtag ticket, would be to define routes in terms of origin and destinations instead of equivalencies.
Requirements to be discussed in the process brainstorm and this issue updated.

PR #315, opened 2017-10-30
SyncBot: Support a directed graph of flows.

sqweelygig:
This to enable the merging of two discourse categories down to one flowdock flow for support.

* Router function supports configuration of which events to route.
* SyncBot configures routers to synchronise threads in particular directions.
* SyncBot reverse-maps the synchronisation of messages along mirrored threads.

Change-Type: patch
Connects-To: #297

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #297


docresources/SyncBot/syncbot-example.sh
```diff
@@ -11,11 +11,10 @@ export PROCBOT_BOTS_TO_LOAD="syncbot"
 export SYNCBOT_PORT="4567"
 export SYNCBOT_NAME="blahbot"
 export SYNCBOT_MAPPINGS="[
-    [
-        { \"service\": \"discourse\", \"flow\": \"blah\" },
-        { \"service\": \"flowdock\", \"flow\": \"blah\" },
-        { \"service\": \"front\", \"flow\": \"blah\" }
-    ]
+    { \"source\": { \"service\": \"discourse\", \"flow\": \"blah\" }, \"destination\": { \"service\": \"flowdock\", \"flow\": \"blah\" } },
```
LucianBuzzo: Some line breaks would go a long way here!

docresources/SyncBot/syncbot-readme.md
```diff
@@ -290,15 +290,17 @@ e.g. `syncbot`
...
+Creation of a thread is synchronised directionally.
 
 e.g.
 ```json
 [
-  [
-    { "service": "blah", "flow": "blah" },
-    { "service": "blah", "flow": "blah" }
-  ]
+  { "source": { "service": "discourse", "flow": "blah" }, "destination": { "service": "flowdock", "flow": "blah" } },
```
LucianBuzzo: Ditto as above ^^

lib/bots/syncbot.ts
```diff
@@ -19,7 +19,7 @@ import * as _ from 'lodash';
 import { ProcBot } from '../framework/procbot';
 import { MessengerService } from '../services/messenger';
 import {
-	BasicMessageInformation, CreateThreadResponse, FlowDefinition,
+	BasicMessageInformation, CreateThreadResponse, DirectedEdgeDefinition, FlowDefinition,
```
LucianBuzzo: NIT: Splitting this import up so there is one import per line would make this a lot more readable

lib/bots/syncbot.ts
```diff
@@ -39,10 +39,11 @@ export class SyncBot extends ProcBot {
 	 * @param to         Definition {service, flow} of the flow being emitted to.
 	 * @param messenger  Service to use to interact with the cloud.
 	 * @param logger     Logger to use to interact with the maintainers.
+	 * @param actions    An array of which actions to sync along this route.
 	 * @returns          Function that processes events from the messenger.
 	 */
 	private static makeRouter(
-		from: FlowDefinition, to: FlowDefinition, messenger: MessengerService, logger: Logger
+		from: FlowDefinition, to: FlowDefinition, messenger: MessengerService, logger: Logger, actions: MessengerAction[],
```
LucianBuzzo: NIT: Splitting this up so that there is one parameter per line would make it a lot more readable

lib/bots/syncbot.ts
```diff
@@ -102,9 +115,16 @@ export class SyncBot extends ProcBot {
...
-					logger.log(LogLevel.DEBUG, `---> Creating thread '${firstLine}' on ${toText}.`);
-					// Create a thread if the quest for connections didn't find any
-					return SyncBot.createThreadAndConnect(to, messenger, data);
+					if (!threadId && _.includes(actions, MessengerAction.CreateThread)) {
+						logger.log(LogLevel.INFO, `---> Creating thread '${firstLine}' on ${toText}.`);
+						// Create a thread if the quest for connections didn't find any
+						return SyncBot.createThreadAndConnect(to, messenger, data);
+					}
+					// Pass on that this has no connected thread
+					return Promise.resolve({
```
hedss: Don't need a `Promise.resolve()` here as you're inside a BlueBird `.then()` clause so the return will automatically default to resolve.

sqweelygig:
Parking this one for a bit, a tweak that's happened in log output has revealed a bug, that is being saved by the API goalkeeper, where it should be saved by the defenders.

lib/services/messenger/translators/translator-scaffold.ts
```diff
@@ -47,30 +47,32 @@ export abstract class TranslatorScaffold implements Translator {
 	 * @param messages  Message to search.
 	 * @param format    Format used to encode the metadata.
 	 */
-	public static extractThreadId(
+	public static extractSource(
 		service: string,
 		messages: string[],
 		format?: MetadataEncoding,
-	): Promise<IdentifyThreadResponse> {
+	): Promise<SourceDescription> {
```
hedss: Is this every likely to be asynchronous? If not, it probably shouldn't return a Promise.

lib/services/messenger/translators/discourse.ts
```diff
@@ -259,7 +259,7 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
 	private static convertReadConnectionResponse(
```
hedss: Does `convertReadConnectionResponse()` ever do anything async? Has this not just pushed the same return issue for a Promise where one's not required up a level?

Also, I'm not very keen on having a multi-line map as a function parameter, I can understand stuff like `_.filter(things, (thing) => !thing`, but not a multi line map.

sqweelygig: Addressing the first point, this conforms to a more general pattern expected when translating a response, which sometimes is asynchronous (specifically translators/front.ts:366).  Issue created.

I've reached out in Flowdock about the second point.

hedss: Cool, just wanted to check whether this could never be asynch, but if it can, it should be Promise based. Have spoken to you about the second item.

sqweelygig: Done.

Issue #113, opened 2017-05-31
SyncBot: Confirm permissions before whispering


sqweelygig:
When someone who isn't allow to whisper attempts this, the post is made public.  This is bad.
https://www.flowdock.com/app/rulemotion/user_happiness/threads/R0gViJ1BfQgcMgr8PpSGKeXQ7Ai

PR #316, opened 2017-10-31
SyncBot: Check permissions before whispering.

sqweelygig:
* When a request to whisper comes in, check the account capabilities first (Discourse natively defaults to public rather than error).

Change-Type: patch
Connects-To: #113

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #113


lib/services/messenger/translators/discourse.ts
```diff
@@ -38,6 +38,32 @@ import { EmitConverters, ResponseConverters, TranslatorErrorCode } from './trans
  */
 export class DiscourseTranslator extends TranslatorScaffold implements Translator {
 	/**
+	 * Converts a provided username, in Discourse format, into a generic username
+	 * @param username  Username to convert
+	 * @returns         Generic equivalent of that username
+	 */
+	public static convertUsernameToGeneric(username: string): string {
```
LucianBuzzo: Can you add unit tests for these username conversion methods?

Issue #203, opened 2017-07-17
SyncBot: Discourse adapter to use communal token


sqweelygig:
The Discourse adapter should use the all-users token available to Discourse administrators.

sqweelygig:
Repurposing this one to reflect the use of the all-users token.

PR #318, opened 2017-11-02
SyncBot: Use Discourse all-users token.

sqweelygig:
* Discourse translator uses statically provided token, assumed to be all-users.
* No longer require the DataHub class for dynamic retrieval of values.
* Allow local development instances of Discourse by making protocol configurable.

Change-Type: patch
Connects-To: #203

---
##### Contributor checklist

- [x] Introduces security considerations
- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #203


lib/services/messenger/translators/discourse.ts
```diff
@@ -309,24 +306,18 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
...
-		.then((token) => {
-			// Pass back details that may be used to connect.
-			return {
-				token,
-				username: convertedUsername,
-				instance: this.connectionDetails.instance,
-				type: ServiceType.Emitter,
-			};
+		// Pass back details that may be used to connect.
+		return Promise.resolve({
```
LucianBuzzo: AFAICT This doesn't need return a promise anymore, since the only place this is used is here https://github.com/resin-io-modules/resin-procbots/blob/2c928043584e70dafc35fee0a8932837e5051c8a/lib/services/messenger.ts#L86 it can just return an object.

sqweelygig:
With my apologies for a premature fixup, have implemented the changes @LucianBuzzo suggested.

PR #326, opened 2017-11-07
SyncBot: Improve debug output of synchronising events

sqweelygig:
Change-Type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


lib/bots/syncbot.ts
```diff
@@ -51,9 +51,10 @@ export class SyncBot extends ProcBot {
 			const sourceText = `${data.source.service} (${data.source.flow})`;
 			const fromText = `${from.service} (${from.flow})`;
 			const toText = `${to.service} (${to.flow})`;
+			const firstLine = data.details.text.split(/[\r\n]/)[0];
```
hedss: Is this definitely right? If you want to split on either an OSX/Linux newline OR a Windows newline, surely you need something like `\r?\n`, else you'll end up with a newline from a Windows input?

sqweelygig: This splits on anything in the set of new line characters, creating an array that might have too many elements in it, and then takes the head of the array.

sqweelygig: Indeed, just noticed I was clever and didn't /g that regex.  It splits on the first thing that is a new line character.

Issue #333, opened 2017-11-14
SyncBot: Connection search to be more discerning


sqweelygig:
There are occasions where message content gets picked up as a connect message, this shouldn't happen.

PR #334, opened 2017-11-14
SyncBot: Improve resilience of thread connections

sqweelygig:
* Pass results of keyword searches through regex filter before returning id

Fixes: https://www.flowdock.com/app/rulemotion/public-s-community/threads/nZIJwAJBeUGEeDT832nipFXyhvs
Change-Type: patch
Connects-To: #333 #342 

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #333


lib/services/messenger/translators/flowdock.ts
```diff
@@ -103,16 +103,25 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
 	 * @returns         Promise that resolve to the thread details.
 	 */
 	private static convertReadConnectionResponse(
-		message: TransmitInformation, response: FlowdockResponse
+		message: TransmitInformation, response: FlowdockMessage[]
 	): Promise<IdentifyThreadResponse> {
-		if (response.length > 0) {
-			const idFinder = new RegExp(`\\[${message.source.service} thread ([\\w\\d-+\\/=]+)]`, 'i');
+		const idFinder = new RegExp(`${message.source.service} thread ([\\w\\d-+\\/=\\_]+)]`, 'i');
```
LucianBuzzo: It looks like this regex may have a trailing `]` character, I'm not sure if this is intentional or not

lib/services/messenger/translators/discourse.ts
```diff
@@ -240,19 +240,29 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
 	 * @returns         Promise that resolves to emit instructions.
 	 */
 	private static convertReadConnectionResponse(
-		message: TransmitInformation, response: DiscourseResponse
+		message: TransmitInformation, response: DiscoursePostSearchResponse
 	): Promise<IdentifyThreadResponse> {
-		const idFinder = new RegExp(`${message.source.service} thread ([\\w\\d-+\\/=]+)`, 'i');
-		if (response.posts.length > 0) {
-			return Promise.resolve({
+		const idFinder = new RegExp(`${message.source.service} thread ([\\w\\d-+\\/=\\_]+)`, 'i');
```
LucianBuzzo: This code seems like it could be generalised and shared between the two classes.

hedss: If this is meant to be the same as the finder in `flowdock.ts`, then this def. needs commonality so errors can't creep in (like the difference that currently exists) as Lucian's pointed out.

lib/services/messenger/translators/discourse.ts
```diff
@@ -240,19 +240,29 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
...
+		// Maps each item to its matched ID, and then removes falsey values.
+		const matches = _.compact(_.map(
+			response.posts,
+			(comment) => {
 				// Because Discourse assumes that when my POST contains:
 				// hyphen hyphen ("--") then what I mean is en dash ("–")
 				// hyphen hyphen hyphen ("---") then what I mean in em dash ("—")
-				thread: response.posts[0].blurb.replace('–', '--').replace('—', '---').match(idFinder)[1],
+				const fixedText = comment.blurb.replace(/–/g, '--').replace(/—/g, '---');
```
hedss: Just so I understand this, the blurb is the post text, right? So what if someone's written:
```
var myVar = 1 - 2;
```
?

If not and this is purely private preamble, obviously it's not a problem.

sqweelygig: @hedss are you making this comment because all the dashes look the same in a fixed width font?   There are actually three different dash characters in that line.

lib/services/messenger/translators/flowdock.ts
sqweelygig: Intentional, but not required.

lib/services/messenger/translators/translator-scaffold.ts
```diff
@@ -84,6 +85,26 @@ export abstract class TranslatorScaffold implements Translator {
 		}
 	}
 
+	protected static extractThreadId(service: string, messages: string[]): Promise<IdentifyThreadResponse> {
```
LucianBuzzo: Lets add some tests for this ;) https://github.com/resin-io-modules/resin-procbots/pull/339

lib/services/messenger/translators/discourse.ts
hedss: Same question applied, if someone uses an endash or emdash in a comment, will it get replaced?

sqweelygig: Yes, it will, however...
1) This replaced text is only used to search for IDs, which have been witnessed containing double hyphens, but not witnessed with en or em dashes.
2) Discourse does this replacement in a destructive, non-reversible manner.

lib/services/messenger/translators/discourse.ts
```diff
@@ -240,20 +243,18 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
...
 				// Because Discourse assumes that when my POST contains:
 				// hyphen hyphen ("--") then what I mean is en dash ("–")
 				// hyphen hyphen hyphen ("---") then what I mean in em dash ("—")
-				thread: response.posts[0].blurb.replace('–', '--').replace('—', '---').match(idFinder)[1],
-			});
-		}
-		return Promise.reject(new TranslatorError(
-			TranslatorErrorCode.ValueNotFound, 'No connected thread found by querying Discourse.'
-		));
+				return comment.cooked.replace(/–/g, '--').replace(/—/g, '---').replace(/&amp;/g, '&');
```
LucianBuzzo: Is it worth refactoring this logic and adding a unit test for it?

Issue #337, opened 2017-11-16
SyncBot: Revoke tag handling


sqweelygig:
The tag handling is buggy and should be revoked.  Needs revokation before reimplementation because it is also messing with sandbox for some reason.

PR #338, opened 2017-11-16
SyncBot: revoke tag handling pending re-implementation

sqweelygig:
Change-Type: patch
Connects-To: #337 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #337


lib/bots/syncbot.ts
```diff
@@ -133,6 +134,11 @@ export class SyncBot extends ProcBot {
 						.return();
 					}
 				});
+				*/
```
pimterry: We should just delete this code entirely, if we want it to be disabled. We've got it in git anyway, and it'll make everything much clearer.

lib/bots/syncbot.ts
```diff
@@ -231,6 +238,7 @@ export class SyncBot extends ProcBot {
 			source: 'syncbot',
 		});
 	}
+	*/
```
pimterry: Ditto

PR #339, opened 2017-11-16
Add mocha unit test framework

LucianBuzzo:
Includes a basic test file, mainly as an example and a jumping-off point

change-type: minor

---
##### Contributor checklist

- [X] Tests are included
- [X] Affects the development, build or deployment processes of the component

##### Reviewer Guidelines
---


tests/services/messenger/translators/discourse.spec.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference types="mocha" />
```
pimterry: You shouldn't need this - I think if you just remove it everything should work without...

LucianBuzzo: That's what I thought as well, but there are some issues with ts-node not picking up `@types` declarations with this combination of TS and mocha.
https://github.com/TypeStrong/ts-node/issues/271
I tried a few different options, and this was the cleanest solution, I'd like to come back and resolve this in the future though.

PR #341, opened 2017-11-21
SyncBot: deployment transparency

sqweelygig:
This is a quick step to make things easier right now, accepting that this will be used to inform a separate repo soon.

* Provide an example variables.yml file for git-backed k8s deployments.
* Updated .sh and .md files, they had got left behind in an update.

Change-Type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


brownjohnf:
I see what you're doing, and I think this is better than nothing, but I think the original proposal was to split out secrets config and non-secrets config, allowing the secrets to be injected as key=<simple value>, and all of the rest is read from the config file in the repo. So something like

```
# variables.yml
syncbot:
        "front_secret": "<<INJECT_front_secret>>"
        "front_token": "<<INJECT_front_token>>"

# syncbot-config.yml
"front":
        "channelPerInbox": <<TODO object of inb_blah to cha_blah here. Exactly the same object structure as previous SYNCBOT_FRONT_CONSTRUCTOR_OBJECT.inbox_channels>>
```

Does that make sense? Heds, is that what you were thinking?

lib/bots/syncbot.ts
```diff
@@ -31,6 +33,15 @@ import { Logger, LogLevel } from '../utils/logger';
  * A bot that mirrors threads across services.
  */
 export class SyncBot extends ProcBot {
+	public static getConfig(key: string): string {
+		if (process.env[`SYNCBOT_${key}`]) {
+			return process.env[`SYNCBOT_${key}`];
+		}
+		const fileContents = fs.readFileSync(`./configs/${process.env.CONFIG_TO_LOAD}.yml`, 'utf8');
+		const varsFromFile = yaml.safeLoad(fileContents);
+		return varsFromFile[`SYNCBOT_${key}`];
```
LucianBuzzo: So this means that if env vars aren't set you fallback to reading values from the config file? That seems a lot like you could introduce faults into your deployment very easily, by forgetting to set an env var. Surely you should be throwing an error here if the configuration is missing?

lib/services/messenger/datahubs/configuration.ts
```diff
@@ -1,54 +0,0 @@
-/*
```
LucianBuzzo: This git commit doesn't provide an explanation for why these files have been deleted, at a surface level it appears to be entirely unrelated to your commit message.

sqweelygig: Reworded commit message.

sqweelygig:
I've had a rethink on the getConfig function, give me a moment.

hedss:
@brownjohnf Yep, that's pretty much what I was thinking, tags that were replaced!

docresources/SyncBot/syncbot-example.sh
```diff
@@ -7,16 +7,9 @@ rm -rf ./build
 gulp build
 
 export PROCBOT_BOTS_TO_LOAD="syncbot"
-
+export CONFIG_TO_LOAD="syncbot_test"
```
hedss: This should be `SYNCBOT_CONFIG_TO_LOAD`, else it doesn't fit into the current configuration file paradigm.

lib/bots/syncbot.ts
```diff
@@ -31,6 +33,20 @@ import { Logger, LogLevel } from '../utils/logger';
  * A bot that mirrors threads across services.
  */
 export class SyncBot extends ProcBot {
+	public static getConfig(key: string): string {
+		if (process.env[`SYNCBOT_${key}`] !== undefined) {
+			return process.env[`SYNCBOT_${key}`];
+		}
+		if (process.env.CONFIG_TO_LOAD) {
+			const fileContents = fs.readFileSync(`./configs/${process.env.CONFIG_TO_LOAD}.yml`, 'utf8');
```
hedss: ProcBots already has a configuration framework (`retrieveConfiguration()`/`processConfiguration()`) that all bots inherit that does all of the heavy lifting for you, why is there extra code here which might become stale?

lib/services/messenger/translators/translator-scaffold.ts
```diff
@@ -36,6 +38,16 @@ export enum MetadataEncoding {
  * and service specific forms.
  */
 export abstract class TranslatorScaffold implements Translator {
+	public static getConfig(key: string): string {
```
hedss: Again, the framework provides the ability to get configurations.

sqweelygig:
Closing this one and re-implementing, hedss has pointed out a pretty fundamental rethink that's required.

Issue #324, opened 2017-11-07
SyncBot: Reduce steps between repo and deployment


sqweelygig:
Devops, in general, shouldn't need to be experts on how a codebase is configured.  SyncBot, at the moment, requires this.  The configuration should be git backed for everything except the tokens.

PR #345, opened 2017-11-23
SyncBot: Put non-secret configuration into repo

sqweelygig:
* Migrated all process.env uses to the top level bot, syncbot
* Created a recursive function to inject secrets into a complex object
* SyncBot config now loaded from a file then injected

Change-Type: minor
Connects-To: #324 

---
##### Contributor checklist

- [x] Tests are included
- [x] Affects the development, build or deployment processes of the component

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #324


hedss:
@sqweelygig Does this no longer connect to #324 then?

lib/framework/procbot.ts
```diff
@@ -44,6 +44,42 @@ interface ServiceMap<T> {
...
+		return value;
+	}
+
+	/**
+	 * A method to inject values from env vars if required
+	 * @param raw  String to translate, use <<INJECT_BLAH>> to retrieve BLAH
+	 * @returns      Environment variable or original string
+	 */
+	private static stringInjector(raw: string): string {
+		if (/^<<INJECT_.*>>$/.test(raw)) {
```
hedss: Seems a strange way of doing this as opposed to just using a `raw.match()`. 

I think it's also make sense to make this a non-greedy repeating match, as otherwise it's fairly limited and wouldn't be able to deal with strings that use injectables within them:

```
creds: '<<INJECT_USERNAME>>/<<INJECT_PASSWORD>'
```

lib/services/messenger/translators/flowdock.ts
```diff
@@ -101,18 +106,20 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
...
-	 * @param response  The response from the SDK.
-	 * @returns         Promise that resolve to the thread details.
+	 * @param metadataConfig  Configuration of how metadata was encoded
+	 * @param message         The initial message that prompted this action.
+	 * @param response        The response from the SDK.
+	 * @returns               Promise that resolve to the thread details.
 	 */
 	private static convertReadConnectionResponse(
-		message: TransmitInformation, response: FlowdockMessage[]
+		metadataConfig: MetadataConfiguration, message: TransmitInformation, response: FlowdockMessage[],
```
hedss: Why the comma on the end?

lib/services/messenger/translators/front.ts
```diff
@@ -193,11 +198,12 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
 	/**
 	 * Converts a provided message object into instructions to create a thread.
 	 * @param connectionDetails  Details of the connection to find things like channels to use.
+	 * @param metadataConfig     Configuration to use to encode metadata
 	 * @param message            object to analyse.
 	 * @returns                  Promise that resolves to emit instructions.
 	 */
 	private static createThreadIntoEmit(
-		connectionDetails: FrontConstructor, message: TransmitInformation
+		connectionDetails: FrontConstructor, metadataConfig: MetadataConfiguration, message: TransmitInformation,
```
hedss: Comma on the end?

lib/services/messenger/translators/front.ts
```diff
@@ -236,11 +247,12 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
 	/**
 	 * Converts a provided message object into instructions to create a message.
 	 * @param connectionDetails  Details of the connection to find things user ID.
+	 * @param metadataConfig     Configuration to use to encode metadata
 	 * @param message            object to analyse.
 	 * @returns                  Promise that resolves to emit instructions.
 	 */
 	private static createMessageIntoEmit(
-		connectionDetails: FrontConstructor, message: TransmitInformation
+		connectionDetails: FrontConstructor, metadataConfig: MetadataConfiguration, message: TransmitInformation,
```
hedss: Comma again, am I missing something obvious here? There are many more functions with last parameter including a comma, I'm not going to point them all out. We certainly allow commas in the final entry in a structure, but not in a function.

lib/framework/procbot.ts
```diff
@@ -44,6 +44,43 @@ interface ServiceMap<T> {
...
+	}
+
+	/**
+	 * A method to inject values from env vars if required
+	 * @param raw  String to translate, use <<INJECT_BLAH>> to retrieve BLAH
+	 * @returns      Environment variable or original string
+	 */
+	private static stringInjector(raw: string): string {
+		let cooked = raw;
+		_.forEach(raw.match(/<<INJECT_.*?>>/g), (match) => {
```
hedss: Hmm, is there any reason why we're not using `//.exec()` here? This is essentially the kind of thing it was designed for:
```typescript
let cooked = raw;
let match = /<<INJECT_(.*?)>>/g.exec(cooked);
while (match) {
    const key = match[1];
    if (typeof process.env[key] === 'string') {
        cooked = cooked.replace(match[0], process.env[key]);
    } else {
        throw new Error(`${key} expected in environment variables, but not found.`);
    }
    match = /<<INJECT_(.*?)>>/g.exec(cooked);

    return cooked;
}
```
It's a lot more readable, means we can capture the key without having to do that ugly set of 'replaces'.

Issue #357, opened 2017-12-06
SyncBot: Should indicate privacy on Flowdock


sqweelygig:
Output: 
```
#reply
public comment text here
```

Input: #reply anywhere in text

PR #359, opened 2017-12-07
SyncBot: Indicate privacy in Flowdock

sqweelygig:
* When encoding a message to flowdock put a #reply in if it's public
* When decoding a message on flowdock look for #reply to indicate publicity
* No longer have any use for char MD formats or indicators

Change-Type: patch
Connects-To: #357 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #357


lib/services/messenger/translators/flowdock.ts
```diff
@@ -42,33 +42,104 @@ import {
...
+	 * @returns        Object of content, genesis and hidden.
+	 */
+	public static extractMetadata(
+		message: string, format: MetadataEncoding, config: MetadataConfiguration,
+	): TranslatorMetadata {
+		const superFormat = (format === MetadataEncoding.Flowdock) ? MetadataEncoding.HiddenMD : format;
+		const metadata = TranslatorScaffold.extractMetadata(message, superFormat, config);
+		const findHashtag = new RegExp(`(?:^|\\W)#${config.publicity.shown}(?:\\W|$)`, 'i');
+		if (format === MetadataEncoding.Flowdock && metadata.genesis === null) {
+			metadata.hidden = !(new RegExp(findHashtag).test(metadata.content));
```
hedss: No need for a new regexp based on the already compiled regexp.

lib/services/messenger/translators/flowdock.ts
```diff
@@ -42,33 +42,104 @@ import {
...
+		}
+		if (metadata.hidden === false) {
+			metadata.content = metadata.content.replace(new RegExp(findHashtag), '').trim();
```
hedss: No need for a new regexp based on the already compiled regexp.

Issue #374, opened 2017-12-20
SyncBot: Restore log redaction


sqweelygig:
Somewhere along the way log redaction stopped being used by SyncBot, and needs to be back again.

sqweelygig:
Could probably use `process.env` values really well here.

PR #375, opened 2018-01-04
374 log redaction

sqweelygig:
* SyncBot: Log the full error response from services.
* ProcBot: Add function to calculate the key/value pairs that would be injected.
* ServiceScaffold: Store logger instance and construction data when constructed.
* ServiceScaffold: Bubble to error handler details of the request and the session.
* Logger: Allow configuration of a list of strings to redact.
* Logger: Move log redaction to the underlying, universal, function to catch more outputs.

Test routine:
--

1) For each of Discourse, Front & Flowdock:
    1) Introduce a deliberate error in the payload delivered to the service
    1) Capture an output of that error occurring on the master branch
    1) Capture an output of that error occurring on the feature branch
    1) Confirm that I can see the error in the second capture
    1) Confirm that secrets are not present in the second capture
1) Remove the captures via their own commit, so they can be cited for now, but disappear on merge

Change-Type: patch
Connects-To: #374
See: https://github.com/resin-io-modules/resin-procbots/commit/a915cbc983d089e3d7ddcabb9ff890859cc37e86

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #374


sqweelygig:
The secrets are controlled by the configuration of the server, not by the people who interact with SyncBot.

lurch:
Pinging @izavits as this might tangentially connect to https://github.com/resin-io/hq/pull/897 ?

lib/framework/procbot.ts
```diff
@@ -58,29 +59,52 @@ export class ProcBot {
 			return _.mapValues(value, ProcBot.injectEnvironmentVariables);
 		}
 		if (_.isString(value)) {
-			return ProcBot.stringInjector(value);
+			return ProcBot.stringInjector(value).cooked;
 		}
 		return value;
 	}
 
+	public static calculateInjections(value: any): Dictionary<string> {
```
hedss: Don't like `calculateInjections` as a name as no actual transformation is occurring. `determineInjections`, perhaps?

lib/utils/logger.ts
```diff
@@ -132,6 +134,12 @@ export class Logger {
 	// Generic output method for either type.
 	private output(level: number, classLevel: number, levelStrings: string[], message: string): void {
 		if (level <= classLevel) {
+			if (!_.isEmpty(this._secrets) && process.env.PROCBOT_LOG_SECRETS !== 'true') {
```
hedss: Were we not going to base this on the output level? I don't really like having a new envvar specifically for logging secrets.

lib/utils/logger.ts
```diff
@@ -58,6 +58,7 @@ export class Logger {
 	// These can be overriden by specific methods.
 	private _logLevel = _.parseInt(process.env.PROCBOT_LOG_LEVEL) || LogLevel.INFO;
 	private _alertLevel = _.parseInt(process.env.PROCBOT_ALERT_LEVEL) || AlertLevel.ERROR;
+	private _secrets: string[] = [];
```
hedss: This doesn't need to be `_secrets` as there's no setter/getter with the same name.

lib/utils/logger.ts
```diff
@@ -104,19 +105,20 @@ export class Logger {
 		this._alertLevel = level;
 	}
 
+	/**
+	 * Add a list of strings to the list of strings to redact.
+	 * @param secrets  Secrets to redact from any output.
+	 */
+	public pushSecrets(secrets: string[]): void {
```
hedss: However, considering the comment above, arguably this should be a setter:
```typescript
public set secrets(...) { ... }
```
In which case, the class member name *should* be `_secrets`. :)

lib/utils/logger.ts
sqweelygig: Note to self: ignored as per next comment.

Issue #382, opened 2018-01-23
SyncBot: `>` not `#reply`


sqweelygig:
https://www.flowdock.com/app/rulemotion/r-process/threads/ReBUfNSgvSNuNw1AJHsvgc6_h-e

PR #385, opened 2018-01-25
SyncBot: Migrate flowdock's reply syntax to `> `.

sqweelygig:
* Move responsibility for spacing to the individual translator.
* Rework flowdock's extraction of metadata to use `/^> /`.
* Rework flowdock's formatting of text:
  * To use `> ` for privacy.
  * To improve readability of the function.
  * To use less whitespace between elements.
  * To improve provision of optional options.
* Parse the gathering of details from the first message (used as thread details) through standard extractMetadata

Change-Type: patch
Connects-To: #382

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #382


lib/services/messenger/translators/flowdock.ts
```diff
@@ -62,12 +62,14 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
 	): TranslatorMetadata {
 		const superFormat = (format === MetadataEncoding.Flowdock) ? MetadataEncoding.HiddenMD : format;
 		const metadata = TranslatorScaffold.extractMetadata(message, superFormat, config);
-		const findHashtag = new RegExp(`(?:^|\\W)#${config.publicity.shown}(?:\\W|$)`, 'i');
+		const findPublic = '^> ';
```
hedss: My only worry with this is that `>Hello here's a response` is going to get missed. Does there need to be a space in it?

Issue #367, opened 2017-12-14
SyncBot: flowdock 400 errors


sqweelygig:
SyncBot has some weird flowdock 400 errors that I don't know what's happening with, and need to.  To dig. example https://app.frontapp.com/open/cnv_ddwc1t

PR #387, opened 2018-01-26
SyncBot: Improve resilience of Front linked thread searching

sqweelygig:
* Increase the uniqueness of each `duff` string
* Fixed a comment I happened to notice was wrong
* Added error messages for Flowdock 5xx errors

This is because increasing the error output revealed that SyncBot is attempting to post messages to a duff thread id on Flowdock

Change-Type: patch
Connects-To: #367 

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #367


package.json
```diff
@@ -46,6 +47,7 @@
     "@types/request-promise": "^4.1.36",
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
+    "fs-readfile-promise": "^3.0.1",
```
hedss: See comment in testfile as to why I don't think this is required.

tests/services/messenger/translators/front.spec.ts
```diff
@@ -1,10 +1,48 @@
 /// <reference types="mocha" />
 import { expect } from 'chai';
 
+import * as readFile from 'fs-readfile-promise';
```
hedss: If you're just using it for reading a file:
```
import * as FS from 'fs';
const readFile = Promise.promisify(FS.readFile);
```
I'm all in favour of utilities, but this seems extremely heavy for just reading a file when an extra line will get you exactly the same results.

Maybe we should be including a utility service for general FS access.

lib/bots/syncbot.ts
```diff
@@ -269,8 +269,8 @@ export class SyncBot extends ProcBot {
 				title: data.details.title,
 			},
 			source: {
-				message: 'duff',
-				thread: 'duff',
+				message: 'duff_SyncBot_createErrorComment_a',
```
hedss: Just a comment currently, but maybe we should add in future something that allows the creation of a unique set of identifier messages, something that texts a text string then appends line number to it?

sqweelygig: Good shout.  Will add an issue.

Issue #325, opened 2017-11-07
SyncBot: Extend service coverage to include GitHub


sqweelygig:
A repo can be consider analogous to a flow.  An issue can be considered analogous to a thread.  SyncBot can therefore sync GitHub with Flowdock.

sqweelygig:
GH Issues - review requests posted in flow?

PR #397, opened 2018-02-07
SyncBot: Synchronise GitHub repos to Flowdock flows

sqweelygig:
* Added GitHub translator which will translate:
  * new issues from GH
  * comments on issues to and from GH
  * events on issues from GH
  * new PRs from GH
  * threaded (not review/code) comments on PRs to and from GH
  * threaded (not review/code) events on PRs from GH
* Tweaked GitHub service so it can share an express instance with the Discourse service.
* Tweaked services to use a common expectation of Express app or port number for ingress.
* Tweaked SyncBot router to better pass errors straight through.
* Tweaked message privacy to allow private as preferred, but not required.
* Added utility method for line endings to translator-scaffold
* Tidied the existing translators (I noticed some oddities while using them as a base for the GH translator)

Change-Type: minor
Connects-To: #325 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #325


lurch:
As this is WIP, I've told VersionBot to shit up for now :wink:

sqweelygig:
Thanks!

lib/services/github-types.d.ts
```diff
@@ -62,8 +63,8 @@ export interface GithubConstructor extends ServiceConstructor {
 export interface GithubListenerConstructor extends GithubConstructor {
 	/** Path the service should listen on for the webhook. */
 	path: string;
-	/** Port the service should listen on for the webhook. */
-	port: number;
+	/** Server or port the service should listen on for the webhook. */
+	server: Express | number;
```
hedss: Needs a better name than `server`, as it could also be port number.

lib/services/github.ts
```diff
@@ -185,19 +185,16 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
 				return false;
 			}
 
-			// Standard Express installation.
-			const app = express();
-			app.use(bodyParser.urlencoded({ extended: true }));
-			app.use(bodyParser.json());
+			const app = this.massageToExpress(listenerConstructor.server);
```
hedss: Needs a comment, as was here previously, to signal intention. Name's not great, how about something like `getExpressInstance`?

lib/services/github.ts
```diff
@@ -185,19 +185,16 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
...
 
 			// Listen for Webhooks on the path specified by the client.
 			app.post(listenerConstructor.path, (req, res) => {
 				const eventType = req.get('x-github-event') || '';
 				const githubSignature = req.get('x-hub-signature') || '';
-				const payload = req.body;
+				const payload = JSON.parse(req.body);
```
hedss: So, this is because now we have SyncBot which wants to be able to pass any content type. Why doesn't the express instance passed by SB instead specifically add a body parser for each type that it's going to process? The Github Service is only guaranteeing to process JSON, and nothing else (as nothing else is valid), so I really don't like the idea of letting it try and parse *anything*. It's essentially passing `any` to the body parser.

lib/services/github.ts
```diff
@@ -575,6 +568,21 @@ export class GithubService extends WorkerClient<string> implements ServiceListen
 				`-H "Accept: ${this.ghApiAccept}" https://api.github.com/`);
 		});
 	}
+
+	private massageToExpress(server: number | express.Express): express.Express {
+		if (_.isNumber(server)) {
+			const app = express();
+			app.use(bodyParser.urlencoded({ extended: true }));
+			app.use(bodyParser.text({type: '*/*'}));
```
hedss: See comment in the app hook.

lib/services/service-scaffold.ts
```diff
@@ -92,7 +92,7 @@ export abstract class ServiceScaffold<T> extends WorkerClient<T> implements Serv
 			// This passes to the inheriting class responsibility for understanding the payload structure.
 			this.expressApp.use(bodyParser.text({type: '*/*'}));
```
hedss: Related to comment in the Github Service, I'd far rather a set of content types and parsers to use were passed in, rather than just `any`.

lib/services/messenger/translators/github.ts
```diff
@@ -0,0 +1,303 @@
...
+	 */
+	private static readConnectionIntoEmit(message: TransmitInformation): Promise<EmitInstructions> {
+		const splitIds = GithubTranslator.splitIds(message.target);
+		const thread = splitIds.number;
+		if (!thread) {
+			return Promise.reject(new TranslatorError(
+				TranslatorErrorCode.IncompleteTransmitInformation, 'Cannot search for connections without a thread.'
+			));
+		}
+		const payload: IssuesGetCommentsParams = _.merge({ per_page: 100, number: thread }, splitIds);
```
hedss: GH Service Emitter handles paging for you. I must be missing something, cause I can't see any handler for if there's more than 100 returned comments.

sqweelygig: Follow me on another journey...
1) `translators/github.ts:137-140` creates a path to an sdk method and a payload, which goes to `messenger.ts:89`
2) `messenger.ts:96-105` bind this to the sdk and bundle it into a context for `messenger.ts:106`
3) `messenger.ts:106` in turn calls the emitter's sendData (in this case `services/github.ts:244`, which deals with pagination.

lib/services/github.ts
sqweelygig: Issue created https://github.com/resin-io-modules/resin-procbots/issues/398

sqweelygig:
No longer WIP!

sqweelygig:
If @brownjohnf could confirm here that he has the required configuration values (GH App ID, GH App webhook secret, GH App private key) then I shall merge this.  I'm holding off because without those INJECTs it would block the deploy pipeline.

sqweelygig:
@brownjohnf Okay, so where in the process should I instruct you how to inject secrets into the environment?
* During the PR to the component?
* During the PR to resin-io/resin-io?
* During the deploy request?
I had assumed the last, since this would match the process to reconfigure the environment, but welcome your advice.

brownjohnf:
yeah, probably during the deploy request is best

PR #403, opened 2018-02-09
SyncBot: Support Sync within the same service

sqweelygig:
The observant in the audience will notice that this doesn't actually configure the OmniFlow.  I want to have this code in production for at least a few days before turning on the firehose.
If SyncBot is ever set up to Sync a flow which is not public, this will reveal information to other Resineers

* Track which flow a message originated from
* Allow messages within the same service, but different flows, to Sync.

Change-Type: patch

---
##### Contributor checklist

- [x] Introduces security considerations
- [ ] Tests are included

##### Reviewer Guidelines
---


tests/services/messenger/translators/discourse.spec.ts
```diff
@@ -2,14 +2,16 @@
 import { expect } from 'chai';
 import * as _ from 'lodash';
 
+import { BasicMessageInformation } from '../../../../lib/services/messenger-types';
```
LucianBuzzo: Using `paths` in you tsconfig can help with these really long file paths --> https://github.com/resin-io-modules/rendition/blob/master/tsconfig.json#L18

sqweelygig: @LucianBuzzo I've just skim read the rendition test files, and note that they all use the relative paths.  I'll be continuing to google and stack overflow about this, but wondered if your brain would be quicker.

package.json
```diff
@@ -61,6 +61,7 @@
     "typescript": "~2.7.1"
   },
   "dependencies": {
+    "@types/bluebird": "^3.5.20",
```
sqweelygig: Note to self: wrong

lib/bots/syncbot.ts
```diff
@@ -128,7 +128,7 @@ export class SyncBot extends ProcBot {
 			if (
 				from.service === data.source.service &&
 				from.flow === data.source.flow &&
-				to.service !== data.details.genesis &&
+				(to.service !== data.details.service || to.flow !== data.details.flow) &&
```
hedss: Ideally parenthesised within.

lib/services/messenger/translators/flowdock.ts
```diff
@@ -63,7 +63,7 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
 		const superFormat = (format === MetadataEncoding.Flowdock) ? MetadataEncoding.HiddenMD : format;
 		const metadata = TranslatorScaffold.extractMetadata(message, superFormat, config);
 		const findPublic = '^> *';
-		if (format === MetadataEncoding.Flowdock && metadata.genesis === null) {
+		if (format === MetadataEncoding.Flowdock && metadata.service === null) {
```
hedss: Parentheses.

PR #409, opened 2018-02-10
ProcBots: Remove TS 2.7.x compiler errors.

hedss:
All valid errors, have fixed appropriately.

This means we can move to latest version of TS. \o/

Connects-to: #408
Change-type: patch
---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #408


lib/framework/procbot.ts
```diff
@@ -308,7 +308,8 @@ export class ProcBot {
 	private addService(type: ServiceType, name: string, data?: ServiceConstructor):
 	ServiceListener | ServiceEmitter | void {
 		// If there's no handle in the constructor data, the handle is the name.
-		let handle = _.get(data, 'handle', name);
+		// We don't use the default in _.get() as TS doesn't imply the type correctly.
```
pimterry: Nit: `imply` should be `infer`

package.json
```diff
@@ -58,7 +58,7 @@
     "ts-node": "^3.3.0",
     "tslint": "^4.5.1",
     "typedoc": "^0.7.1",
-    "typescript": "~2.4.0"
+    "typescript": "^2.7.0"
```
pimterry: This'll work, but if I were you I'd stay pinned a least to a minor version (`~2.7.0`). TS doesn't really follow semver, so bumps can do large breaking things, and this significantly increases the odds this project will randomly stop building in future.

lib/services/service-scaffold.ts
```diff
@@ -252,7 +252,9 @@ export abstract class ServiceScaffold<T> extends WorkerClient<T> implements Serv
 	 * Get the service name, as required by the framework.
 	 * @return  Name of the service.
 	 */
-	abstract get serviceName(): string
+	get serviceName(): string {
+		return this.serviceName;
+	}
```
sqweelygig: 1) Creates an infinite loop..
2) Does not insist that the variable is defined.
3) Would be nice to remove overrides of this that follow the pattern.

sqweelygig: Something like https://github.com/resin-io-modules/resin-procbots/commit/c71844cd75ec59b6e4f43bfc9617b4413cdecb3d

hedss: @sqweelygig Actually, having now had a closer look at this, I really don't like two things:

1. The reason this was changed is because the ServiceScaffold constuctor was trying to use `serviceName`, which is strictly prohibited (this is an abstract property, and therefore invalid)
2. The better way to do this is to pass `serviceName` into the `ServiceScaffold` constructor if we want to do this, and explicitly set it there so it can be used and called by the implementer, removing the private instances from each Service.

I do not have anymore time to do 2. I'm backing out the change I made, and commenting out the logging in the `ServiceScaffold` constructor. If you'd like to make the relevant changes on this PR, that would be helpful.

package.json
hedss: Whilst true about semver, I'm kinda of the opinion we should be 'keeping up with latest TS', and fixing issues as they come up. One of the reasons this languished so long was the work wasn't done to bring it up, and it just kept lapsing. :-/

I'll lock it, but we need to keep on top of version changes better than we have been.

Issue #259, opened 2017-08-29
SyncBot: Hashtag to create a pair of threads


sqweelygig:
SyncBot should take over the Hubot task of representing hashtagged conversations in Front.  When a conversation becomes hashtagged this should create a mirrored conversation in Front and from then on keep them synchronised.

sqweelygig:
Okay, so some notes to self:

Could SyncBot be simply an implementation of the following simple rules:
1) If a comment happens in a monitored flow/category/inbox then create a connected thread/conversation/topic.
2) If a comment happens that includes a hashtag then create a connected thread/conversation/topic.
3) If a comment happens in a connected thread/conversation/topic then synchronise it.

PR #432, opened 2018-03-01
SyncBot: Support for hashtags to create synced conversations

sqweelygig:
* Created a hashtag translator that inherits from Flowdock translator, but:
  * turns events on hashtagged threads into messages
  * uses ...flow.../threads/...thread... as its thread ID
    (the configured part of the route is hashtag, the dynamic part is flow.thread)

* Added a check for if the event is on the destination of the route (and re-presented the logic)
* Fixed a type inheritance typo in the FlowdockEvent object
* Allowed the Flowdock service to receive a different name from child classes
* Built a shim hashtag service as a child of the Flowdock service
* Made various details from the Flowdock translator available to child classes

Change-Type: minor
Connects-To: #259

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #259


lib/services/messenger/translators/flowdock.ts
```diff
@@ -521,6 +500,71 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
...
+	 * @param event  Event to scrutinise for user details
+	 * @returns      Promise that resolves to the alias presented with the message
+	 */
+	protected fetchAlias(event: FlowdockEvent): Promise<string> {
+		if (event.rawEvent.external_user_name) {
+			return event.rawEvent.external_user_name;
+		}
+		const url = `/organizations/${this.organization}/users/${event.rawEvent.user}`;
+		return FlowdockTranslator.fetchFromSession(this.session, url)
+		.then((user) => {
```
hedss: `.get('nick');` could be used instead.

sqweelygig: Wow, that is an incredible function, and one I will use again!

Issue #474, opened 2018-04-12
SyncBot: Add Jellyfish service


sqweelygig:
Create a service/translator set that interacts with a jellyfish instance.

sqweelygig:
Current status, basic functionality present.
To do, go through the MessengerAction list and ensure they are all implemented.

sqweelygig:
Parked because of the deprecation of ProcBots in favour of Concourse & JellySync

PR #479, opened 2018-04-16
SyncBot: Add timestamp to message details

sqweelygig:
This is done as one step of preparation towards putting Jellyfish in as an endpoint

* Add time stamps to received messages

Change-Type: patch
Connects-To: #474

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #474


lib/services/messenger/translators/github.ts
```diff
@@ -253,6 +253,7 @@ export class GithubTranslator extends TranslatorScaffold implements Translator {
 					hidden: _.isSet(metadata.hidden) ? metadata.hidden : false,
 					tags: thread.labels,
 					text: TranslatorScaffold.unixifyNewLines(text),
+					time: event.cookedEvent.data.created_at,
```
hedss: Use `updated_at` rather than `created_at`. Any event that's updated will end up using the original creation date here, which may be far in the past.

tests/services/messenger/translators/discourse.spec.ts
```diff
@@ -1,9 +1,8 @@
 /// <reference types="mocha" />
 import { expect } from 'chai';
 import * as _ from 'lodash';
-
+import * as crypto from 'crypto';
```
hedss: Why is this required? Can't see any changes in the code here that need it?

tests/services/messenger/translators/discourse.spec.ts
```diff
@@ -16,6 +15,7 @@ describe('lib/services/messenger/translators/discourse.ts', () => {
 				hidden: false,
 				tags: [],
 				text: 'cde @test',
+				time: '2018-04-16T12:45:46+00:00',
```
hedss: Quick question on time here. Do timestamps get translated? How are these handled for timezone, or do they all get converted to Zulu?

sqweelygig: In Jellyfish these are stored as an ISO string
In the Jellyfish translator in intend on doing a `moment(time).toISOString()`
Therefore in the common message type these should be in any format that momentjs can parse.
I'm adding this to the interface declaration.

tests/services/messenger/translators/discourse.spec.ts
hedss: My mistake, I hadn't seen it had moved from further down! :)

Issue #481, opened 2018-04-18
SyncBot: Would be nice to interrogate package.json


sqweelygig:
... in a similar way to process and system

PR #482, opened 2018-04-18
SyncBot: Add ability to query the package.json file

sqweelygig:
* Extend matchEnvironment to filter on package.json contents if required
* Extend queryEnvironment to interrogate package.json contents if sought
* Both of these now use Promises because of async file reads
* SyncBot now compiles a package query alongside system and process

Change-Type: patch
Change-Type: major|minor|patch 
Connects-To: #481 

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #481


lib/framework/procbot.ts
```diff
@@ -55,43 +55,75 @@ export class ProcBot {
...
 		const matchSystem = _.every(filter.system, (term) => {
 			return JSON.stringify(os[term.property]) === term.value;
 		});
-		return matchProcess && matchSystem;
+		if (!matchProcess || !matchSystem) {
+			return Promise.resolve(false);
+		} else if (_.isEmpty(filter.package)) {
+			return Promise.resolve(true);
+		}
+		return fsReadFile('./package.json')
```
hedss: If for some reason the bot isn't run from the root of `resin-procbots`, this has the potential to throw, and the entire framework stops (if not called from within a Promise chain with a catch block).

You probably want to use `require.main` to get the entry point and infer the location of the `package.json` from there. AFAIK we don't run via any sort of package manager atm, so this should be safe.

You also need to add a catch which logs if for some reason the file isn't present or parseable or is run from outside a Promise chain.

lib/framework/procbot-types.d.ts
```diff
@@ -54,14 +54,21 @@ export type SystemProperty = {
 	value: any;
 };
 
+export type PackageProperty = {
+	property: string;
+	value: any;
```
hedss: This means that any nested keypair is only going to show up as `[object Object]` potentially in any output. Maybe not an issue, but to do it properly it should really be:
```typescript
value: PackageProperty | any;
```
along with relevant code to descend nested keypairs.

hedss: And actually, I think it's more than that, I think it can be arrays of keypairs too, thinking about it.

lib/framework/procbot.ts
sqweelygig: @hedss re: root.  As far as I can tell `npm start` puts the cwd to the location of package.json

lib/framework/procbot-types.d.ts
sqweelygig: https://www.flowdock.com/app/rulemotion/testing/threads/qHdUgh9StUKyNsTeyHKY0pHl-4o

sqweelygig:
Note to self: Have requested a quick chat.

lib/framework/procbot.ts
hedss: @sqweelygig Yes, it does. But process managers (such as `pm2`) abstract the CWD, and there's absolutely no telling how we decide to start the process in the future, esp. moving forwards with K8. :)

sqweelygig:
https://www.flowdock.com/app/rulemotion/d-bots/threads/L4xuHkpbapL--3pHhUTUTu6GS9e

lib/framework/procbot-types.d.ts
sqweelygig: Note to self: this feedback replaced with feedback to use util.inspect.

sqweelygig: PPS: done

Issue #470, opened 2018-04-09
SyncBot: Add E2E tests


sqweelygig:
For example test https://github.com/resin-io-modules/resin-procbots/pull/468
We should be doing a set of end to end testing to catch changes to services, and bugs on a grander perspective than unit tests.

PR #493, opened 2018-04-27
SyncBot: Add end to end test suite for SyncBot

sqweelygig:
Just a little thing, but I've added an End To End test suite to SyncBot.

Change-Type: patch
Connects-To: #470

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #470


lib/services/messenger/translators/discourse.ts
```diff
@@ -328,12 +347,33 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
 		return Promise.resolve({});
 	}
 
+	private static convertReadThreadResponse(
+		config: MetadataConfiguration,
+		show: 'reply' | 'whisper' | 'all',
+		_message: TransmitInformation,
+		response: any,
+	): Promise<TranslatorMetadata[]> {
+		const messages = _.compact(_.map(response.post_stream.posts, (comment) => {
```
hedss: May as well just:
```typescript
return Promise.resolve(_.compact...);
```
I don't think it's messy, given it's a single function within the resolve.

lib/services/messenger/translators/flowdock.ts
```diff
@@ -383,27 +384,34 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
...
+	private static readThreadIntoEmit(message: TransmitInformation, term?: string): Promise<FlowdockEmitInstructions> {
 		// Check we have a thread.
 		const threadId = message.target.thread;
 		if (!threadId) {
 			return Promise.reject(new TranslatorError(
-				TranslatorErrorCode.IncompleteTransmitInformation, 'Cannot search for connections without a thread.'
+				TranslatorErrorCode.IncompleteTransmitInformation, 'Cannot read a thread without a thread.'
 			));
 		}
+		const payload: Dictionary<string> = {
```
hedss: Just a comment, but given the object property `payload` being returned isn't actually the same `payload` being defined, maybe `body`?

Else I'd personally be expecting:
```typescript
return Promise.resolve({ method: ['get'], payload });
```

lib/services/messenger/translators/flowdock.ts
```diff
@@ -416,12 +424,33 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
 		return Promise.resolve(_.map(response, (comment) => comment.content));
 	}
 
+	private static convertReadThreadResponse(
+		config: MetadataConfiguration,
+		show: 'reply' | 'whisper' | 'all',
+		_message: TransmitInformation,
+		response: FlowdockMessage[],
+	): Promise<TranslatorMetadata[]> {
+		const messages = _.compact(_.map(response, (comment) => {
```
hedss:
```typescript
return Promise.resolve(_.compact());
```

lib/services/messenger/translators/front.ts
```diff
@@ -127,6 +128,38 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
...
+	 * Converts a response into the generic format.
+	 * @param _message  The initial message that prompted this action.
+	 * @param response  The response from the SDK.
+	 * @returns         Promise that resolve to the thread details.
+	 */
+	public static convertListWhispersResponse(
+		metadataConfig: MetadataConfiguration, _message: TransmitInformation, response: ConversationComments
+	): Promise<TranslatorMetadata[]> {
+		const rawComments = _.reverse(_.map(response._results, (comment) => comment.body ));
+		const comments = _.map(rawComments, (commentObject) => {
```
hedss:
```typescript
return Promise.resolve(_.map());
```

lib/services/messenger/translators/front.ts
```diff
@@ -127,6 +128,38 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
...
+	 * Converts a response into the generic format.
+	 * @param _message  The initial message that prompted this action.
+	 * @param response  The response from the SDK.
+	 * @returns         Promise that resolve to the thread details.
+	 */
+	public static convertListRepliesResponse(
+		metadataConfig: MetadataConfiguration, _message: TransmitInformation, response: ConversationComments
+	): Promise<TranslatorMetadata[]> {
+		const rawComments = _.reverse(_.map(response._results, (comment) => comment.body ));
+		const comments = _.map(rawComments, (commentObject) => {
```
hedss:
```typescript
return Promise.resolve(_.map());
```
And same for any other instances in the code which make creating an extra variable probably unnecessary.

package.json
```diff
@@ -26,6 +26,7 @@
   },
   "scripts": {
     "build": "gulp build",
+    "full-test": "nyc mocha -r ts-node/register tests/**/*.spec.ts tests/**/**/*.spec.ts tests/**/**/**/*.spec.ts tests/**/*.e2e.ts tests/**/**/*.e2e.ts tests/**/**/**/*.e2e.ts",
```
hedss: Unless I'm missing something, you only need to glob once for each pattern as `/**/` with `-r` says 'recursively descend the tree matching all suffix patterns`:
```json
"full-test": "nyc mocha -r ts-node/register tests/**/*.spec.ts tests/**/*.e2e.ts"
```

package.json
```diff
@@ -26,6 +26,7 @@
...
     "start": "node ./bin/procbots -b ${PROCBOT_BOTS_TO_LOAD}",
     "test": "nyc mocha -r ts-node/register tests/**/*.spec.ts tests/**/**/*.spec.ts tests/**/**/**/*.spec.ts"
```
hedss: Same as above goes for here, as well. You should only need:
```json
"test": "nyc mocha -r ts-node/register tests/**/*.spec.ts"
```

package.json
sqweelygig: Prior wisdom has me think this is not the case, will make this citable.

sqweelygig: tests2
tests

package.json
sqweelygig: @hedss, see above

package.json
hedss: This confused me for a couple of minutes, because I couldn't see very well what was going on. A short text explanation would have done so much more for me.

So, the standard pattern for globbing in the shell is to quote them. I should have double-checked what I did instead of C&Ping. This should work fine for example:

```
nyc mocha -r ts-node/register 'tests/**/*.spec.ts'
```

package.json
hedss: @sqweelygig See above above. :)

Issue #501, opened 2018-05-07
SyncBot: Investigate Front render


sqweelygig:
**@sqweelygig hey -- messages aren't rendered properly in Front -- this one is unreadable. Same thing applies to the subject, where it has escaped HTML entities. Do we miss an intermediate MD rendering step before handing the message over to Front?**

https://www.flowdock.com/app/rulemotion/public-s-community/threads/X67QVLugRjhx4Nc8Yzs7ibxR_-I

sqweelygig:
Code blocks in message content get gobble-de-gooked
Quote marks in message title get gobble-de-gooked

PR #506, opened 2018-05-07
SyncBot: Clean Front render

sqweelygig:
* Add additional debug points to SyncBot to show full message.
* Revert the unnecessary html escaping that marked introduces.
* Propagate the cleaning of markdown onto first message (taken from reply logic) on Front.
* Added tests for above.

Change-Type: patch
Connects-To: #501

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #501


lib/services/messenger/translators/flowdock.ts
```diff
@@ -87,6 +88,7 @@ export class FlowdockTranslator extends TranslatorScaffold implements Translator
 			metadata.content = metadata.content.replace(new RegExp(findPublic, 'igm'), '').trim();
 			if (metadata.title) {
 				metadata.title = metadata.title.replace(new RegExp(findPublic, 'igm'), '').trim();
+				metadata.title = new AllHtmlEntities().decode(metadata.title);
```
dfunckt: Maybe assign an instance to a module-local variable and reuse it to save the instantiation on each request/call?

lib/services/messenger/translators/front.ts
```diff
@@ -335,7 +335,7 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
 			const messageString = TranslatorScaffold.convertPings(message.details.text, converter);
 			const createThreadData: MessageRequest.Send = {
 				author_id: userId,
-				body: `${messageString}<br />${metadataString}`,
+				body: `${marked(messageString.replace(/\n+/g, '\n\n'))}<br />${metadataString}`,
```
dfunckt: This is slightly nuclear :) “messageString” can potentially include preformatted text that should stay intact and this blindly replaces newlines in it. I think leaving it to “marked” is a better option.

lib/bots/syncbot.ts
```diff
@@ -286,6 +287,7 @@ export class SyncBot extends ProcBot {
 					}
 					if (!threadId && _.includes(actions, MessengerAction.CreateThread)) {
 						logger.log(LogLevel.INFO, `---> Creating thread '${data.details.title}' on ${toText}.`);
+						logger.log(LogLevel.DEBUG, `---> Full message from above, '${data.details.text}'.`);
```
dfunckt: Not sure if this logger is some external library, but it might worth defining info/debug/etc methods on it and invoke them directly instead of needing to reach out to another type (i.e. LogLevel) — it feels slightly cumbersome.

sqweelygig: Logger, and by extension LogLevel, are an internal utility class that really could be extended with this structure.  But!  It is used all over ProcBots, so would not be a trivial amount of work.  I therefore have created an issue to contain this thought.

dfunckt: Yeah, it need not happen overnight. You might as well add the methods, start using them from now on, and whenever you hit upon code that passes LogLevel manually, change that too.

dfunckt: (And by that I don't mean those methods should be added in *this* PR :)

sqweelygig: Fair enough, that does tone it down into a trivial enough amount of work, and is happening right now!

Issue #547, opened 2018-06-13
SyncBot: Images don't quite translate right


sqweelygig:
Placeholder issue for now, so I don't forget it.

sqweelygig:
https://www.flowdock.com/app/rulemotion/public-s-community/threads/NxVJuPwKPi6NyRpEjjLTrTjUc9A

sqweelygig:
Taken from my inbox nags:
Confer with Juan that images cannot be muddled out of Front

sqweelygig:
There is the expectation that images provided in a comment will work across all platforms.  This needs testing, the holes found, patched, etc.

PR #549, opened 2018-06-14
SyncBot: Fix image url translation

sqweelygig:
* Prepend server details when discourse translates an image url
* Prepend flowdock path when flowdock translates a file
* Cannot perform similar translation for Front, because Front
  * Seriously, their UI uses a different version of the API to everyone else, and everyone else doesn't have access to attachments.

Change-Type: patch
Connects-To: #547

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #547

### DIFF
--- a/lib/services/messenger/translators/discourse.ts
+++ b/lib/services/messenger/translators/discourse.ts
@@ -466,7 +466,14 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
 				if (!imgAlt) {
 					return fullMatch;
 				}
-				return images[imgAlt[1]];
+				// /uploads/this.png is relative and does not have protocol
+				// http://example.com/uploads/this.png is not relative and has protocol
+				// //localhost/uploads/this.png is not relative and does not have protocol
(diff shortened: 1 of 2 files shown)


sqweelygig:
This translates the image urls that Discourse emits into fully qualified links, which then render as images in the Flowdock UI, and as a link to the image in the Front UI.  This does not alter the emitting of images provided through the Front or Flowdock UI's; but these workflows need to be checked, so I'm not going to merge yet.

sqweelygig:
I've put this PR in for re-review because I've now investigated Front and Flowdock, with mixed success.

lekkas:
@sqweelygig so what is the current state? What does mixed success mean?

sqweelygig:
@lekkas sorry, hadn't noticed that the rewording of the commit hadn't been bubbled up.  The mixed success is "Front sucks".

sqweelygig:
This needs revisiting.

Issue #540, opened 2018-05-23
SyncBot: Using hashtags in the discourse message body produces the equivalent Front tags


lekkas:
See https://app.frontapp.com/open/cnv_soyu97 for an example (`24-ubuntu`)

Tags in Front are used/created based on our process needs and should not be created from hashtags in conversations, so we should disable this action.

PR #550, opened 2018-06-14
SyncBot: Do not propagate flowdock tags

sqweelygig:
Flowdock, in general, is quite enthusiastic about tagging messages.
Tags, in Front, have very specific meanings and should be controlled

* No longer bring the tags out of Flowdock

Change-Type: patch
Connects-To: #540

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #540


sqweelygig:
The title is accurate.  This does limit the propagation of tags out of Flowdock **in general** and not just **to Front**, this is because there isn't really a translation of Flowdock to Front, it is a translation of Flowdock to Generic to Front.

lekkas:
ack @sqweelygig - what is 'Generic' though?

sqweelygig:
Generic is a pidgin (https://en.wikipedia.org/wiki/Pidgin) form of chat messages that SyncBot operates on, that each incoming and outgoing payload is translated to and from and attempts to represent the features we want from the services.
One of the main things I am looking to simplify in the rebuilding under Jellyfish is a closer similarity between the data structure of the hub service (currently flowdock) and this pidgin, because Flowdock is a long way, conceptually, from this; and that is the cause of a lot of the SyncBot complication.

Issue #562, opened 2018-07-10
SyncBot: Support threads that shift between synced inboxes


sqweelygig:
When a thread moves from one synced inbox to another, SyncBot incorrectly finds the association with the previous inbox, and nothing then syncs.
The search for a context should be inbox specific.

PR #563, opened 2018-07-11
SyncBot: Make thread searching flow specific

sqweelygig:
There is a failure case where a thread is moved, on Front, from one synced inbox to another.  When new events happen on this thread it finds the existing context, but attempts to send to a frankenstein context of the new flow but the old thread.

* Pass the full flow context up the inheritance to places that search for syncs
* Make the searching for a thread specific to the flow context
  * The regex fallback scrutinises based on flow citation, if present
  * The metadata searcher checks for equality of flow values
* While I was there masked my dev config
* While I was there improved the message sent when a thread moves

Change-Type: patch
Connects-To: #562

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #562


lib/services/messenger/translators/front.ts
```diff
@@ -589,7 +589,7 @@ export class FrontTranslator extends TranslatorScaffold implements Translator {
 			switch (details.event.type) {
 				// If this is an edit, then create a murmur to that effect
 				case 'move':
-					metadata = TranslatorScaffold.emptyMetadata(`This thread was edited by ${details.author}.`);
+					metadata = TranslatorScaffold.emptyMetadata(`This thread was ${details.event.type}d by ${details.author}.`);
```
hedss: NIT: This is going to be whacky English coming out depending on the event type. How about `Thread alteration of type ${details.event.type} by ${details.author}.`?

sqweelygig: Correct, and when we encounter such things I expect them to be covered by the `switch case`, I'll comment to that effect.

Issue #558, opened 2018-06-20
SyncBot: Discourse to fall back upon SyncBot acct


sqweelygig:
When a whisper fails the Discourse service should fall back upon the SyncBot account.

Issue #518, opened 2018-05-14
SyncBot: ensure a frequently occurring Front issue is dealt with well


sqweelygig:
https://www.frontstatus.com/incidents/shl35yl8c3vm
This has happened a few times now, search indexes go slow.  I will code a workaround and good report.

PR #566, opened 2018-07-13
SyncBot: Emit discourse whispers under the syncbot account

sqweelygig:
This is to stem the tide of Resineers that we need to promote to Discourse moderators, which is approaching an expensive threshold.

* Toggle the account used based on whether the message is a whisper
* Prefix whisper text sent with the context of who whispered
* Strip whisper text received of the context of who whispered
* No longer check for and reject non-moderator whispers

Change-Type: patch
Connects-to: #558
Signed-off-by: Andrew Lucas <andrewl@resin.io>

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #558
---- Autogenerated Waffleboard Connection: Connects to #568
---- Autogenerated Waffleboard Connection: Connects to #518


lib/services/messenger/translators/discourse.ts
```diff
@@ -100,6 +100,10 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
 	public static bundleMessage(
 		message: BasicMessageInformation, thread: string, metadataConfig: MetadataConfiguration
 	): Promise<DiscourseEmitInstructions> {
+		if (message.details.hidden) {
+			const whisperPrefix = `${message.details.handle} whispered:\n`;
+			message.details.text = `${whisperPrefix}${message.details.text}`;
```
hedss: Would this all fit in a single line less than 120 chars?

lib/services/messenger/translators/discourse.ts
```diff
@@ -356,6 +339,9 @@ export class DiscourseTranslator extends TranslatorScaffold implements Translato
 	): Promise<TranslatorMetadata[]> {
 		return Promise.resolve(_.compact(_.map(response.post_stream.posts, (comment) => {
 			const messageObject = DiscourseTranslator.extractMetadata(comment.cooked, MetadataEncoding.HiddenHTML, config);
+			if (messageObject.hidden || comment.type === 4) {
```
hedss: Hardcoded value for `comment.type`? Are there any enumerated types we could use?

sqweelygig: Not that discourse publish, I've added a comment to this effect to make it more obvious why it's a magic number.

Issue #564, opened 2018-07-13
SyncBot: Front -> Flowdock to represent meta information


sqweelygig:
Information about whether a ticket is closed should be present as flowdock first message tags.
https://www.flowdock.com/app/rulemotion/d-bots/threads/hRBwt7McJ9Jl2ynZ3Mclh7XFzjg

sqweelygig:
This is going to require that SyncBot sync things that are not messages, which is going to be tricky.

PR #574, opened 2018-07-26
SyncBot: Synchronise Front conversation states into Flowdock

sqweelygig:
- Retype message information into event information to support a broader scope of events
- Add a dictionary of state flags to set/clear to the event information
- Add handling for thread events that specify a change of state
- Pass change of state on Front through to SyncBot
- Add instructions for new Front events to emit
- Store a thread's state in Flowdock as initial message tags
- Explicitly require message and details to encode when stringifying metadata
- Make log line fall back onto thread title
- Put intercomHack into the property pertaining to the current service, rather than the payload
- While I was there my IDE told the Front Session could be readonly and that a regex had redundant escape characters

Change-type: patch
Signed-off-by: Andrew Lucas <andrewl@resin.io>
Connects-to: #564

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---
---- Autogenerated Waffleboard Connection: Connects to #564


sqweelygig:
@lekkas is this a technical as well as UX approve?

lekkas:
This is only a UX approval, as far as I can see there are no UX changes other than the tags, is that correct?

sqweelygig:
Ack to nature of review, and correct to nature of changes.

lib/services/messenger-types.d.ts
```diff
@@ -42,32 +43,47 @@ export interface ReceiptIds extends MessengerBaseIds {
 	message: string;
 	/** The ID of the thread. */
 	thread: string;
+	// https://github.com/resin-io-modules/resin-procbots/issues/301
+	/**
+	 * For Intercom messages on Front, whether they had an empty subject line
+	 * An empty subject line indicates the message was inputted via the Front or Intercom UI
+	 * A subject line indicates the message was inputted via the Front API
+	 * Needed because Front mangles the message into plaintext and refuses to let you see raw
```
hedss: Sigh. :(

sqweelygig: It'll be better in Jellyfish.  At the moment I'm integrating quirky with quirky via quirky, which is a big pain.  Under Jellyfish I'll at least be integrating quirky with quirky via **something we control**.

PR #616, opened 2018-11-15
flowdockhashtag: Fix .match is not a function critical error

jviotti:
We found this because of this error: `TypeError: message.content.match
is not a function`.

The problem is that in various event types, `content` is not a string,
but an object. See https://www.flowdock.com/api/message-types for
examples.

We believe this ended up causing some messages to be lost when syncing
from Front to Flowdock.

Change-type: patch


lekkas:
Instances of the error: 

https://logentries.com/app/5915e005#/search/log/cbccfcbe?log_q=where(%2FCRITICAL%2F)&last=Last%2012%20Hours

PR #634, opened 2018-12-18
Enable channels for OpenBalena and Hello Front inboxes

lekkas:


lekkas:
@hedss @jviotti in a similar fashion, merged right away to accelerate deploying fixes for these inboxes - the missing configuration is most likely the reason why the recently enabled syncs were not properly functioning

PR #654, opened 2020-02-13
Stop syncing outage to support_meta

mbalamat:


mbalamat:
Actually yes. But will do in another PR.

PR #658, opened 2020-06-11
syncbot_prod: Sync user feedback, front <-> flowdock

mbalamat:


mbalamat:
@balena-ci I self-certify!

PR #659, opened 2020-06-17
syncbot_prod: Add channel for U/Feedback

mbalamat:


mbalamat:
@balena-ci I self-certify!

PR #660, opened 2020-07-22
configs/syncbot_prod: Add distribution channel

mbalamat:


mbalamat:
@balena-ci I self-certify!

PR #661, opened 2020-07-29
configs/syncbot_prod: Change distribution flow

mbalamat:


mbalamat:
@balena-ci I self-certify!

PR #662, opened 2020-08-12
configs/syncbot_prod: Stop syncing paid support

mbalamat:


mbalamat:
@balena-ci I self-certify!

PR #663, opened 2020-12-09
lib/services/messenger/translators/front: Fix fetchContactName

mbalamat:
fetchContactName fails because the front-sdk's method getFromLink
malforms the URL it uses to request the contact from Front.
See here https://github.com/balena-io-modules/front-sdk/blob/800205025e8ff8946fbd0908ec1003afdabbc9e4/lib/index.ts#L434
The malformed URL looks like
https://api2.frontapp.com/https://resin.io.api.frontapp.com/contacts/<contact_id>
instead it should be
https://api2.frontapp.com/contacts/<contact_id>


mbalamat:
@balena-ci I self-certify!

PR #664, opened 2021-06-17
Improve error reporting when a listener method throws

LucianBuzzo:
This change serializes errors thrown from listener methods, preserving
additional data such as stack traces, which are very useful for
debugging problems.

Change-type: patch


nazrhom:
@balena-ci retest

LucianBuzzo:
@balena-ci I self-certify!

PR #665, opened 2021-06-18
Rename resin to balena in package.json and README

LucianBuzzo:
Change-type: patch


LucianBuzzo:
@balena-ci I self-certify!

PR #666, opened 2021-06-22
Improve parsing of Front URLs

LucianBuzzo:
This fixes a bug that would occur if Front change the subdomain prefix
on links. 😒

Change-type: patch


LucianBuzzo:
@balena-ci I self-certify!

